PR #8, opened 2012-07-21
Easier-to-use editor links

tgr:
- adds ability to define the format of the editor links as a simple string, without the need to override pathDisplayCallback (and have to deal with stuff like resource files in trace)
- takes default from xdebug.file_link_format if available
- adds example scripts to register custom protocol and open any editor from command line in Windows

Fixes issues [28](http://code.google.com/p/kint/issues/detail?id=28) and [34](http://code.google.com/p/kint/issues/detail?id=34) on the old tracker.

### DIFF
--- a/Kint.class.php
+++ b/Kint.class.php
@@ -321,8 +321,9 @@ class Kint
 		}
 
 		$url = str_replace( array( '%f', '%l' ), array( $file, $line), self::$fileLinkFormat );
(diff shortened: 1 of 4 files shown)


raveren:
Wow, awesome, I'll definitely merge the excellent changes and credit you in  the readme (is that how this works?)!

Sorry for the lack of updates, I continuously improve kint, but it currently has some style issues I'm struggling to fix, so I'm not committing to github yet.

PR #13, opened 2012-08-08
Added composer.json

carlos-granados:
Hi

I added a composer.json file to your (excellent) library. With this it will be possible to install your library using composer (the best dependency and library manager for php). I filled in the details but feel free to change anything like author, description, etc...

Regarding requirements, I don´t know if you require any specific version of PHP and if you require any other library to be installed, please modify the composer.json file to reflect this

Once you have included the composer.json file in your project, you should upload it to packagist.org so that other composer users can install it. The procedure is very simple but if you need any help, don´t hesitate to ask.

Regards and keep up the good work

Carlos

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,24 @@
+{
+    "name": "raveren/kint",
+    "description": "Kint - debugging helper for PHP developers",
+    "keywords": ["kint", "php", "debug"],
+    "type": "library",
+    "homepage": "https://github.com/raveren/kint",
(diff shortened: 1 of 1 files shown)


raveren:
Hey, thanks a lot, I really appreciate it. 

I keep kint PHP5.2-compatible to appeal to users even though namespaces and closures would be really helpful, and kint has no dependencies either. I'll merge the pull request and credit you with the next revision.

Thanks again!

PR #14, opened 2012-08-08
change case of splfileinfo.php

carlos-granados:
Hi,

You incorrectly had this file with the name splFileInfo.php. Since your autoloader converts everything to lower case before loading the files, this made loading of this file fail. I guess you were running in a system where file names are not case sensitive like Windows and that is why you probably didn´t notice

I renamed the file to lowercase

Carlos


raveren:
Hi, I am very well aware of case sensitive autoloading issues, and have definitely renamed the culprit, no idea how it got camelCased again..

Thanks!

carlos-granados:
Hi

Packagist by default only updates packages once a day, so these changes are
not yet reflected on the packagist repository.

If you want packagist to update packages as soon as you commit new changes
(this is great for developers who follow your product and who might need a
bug fix), you need to set up a github hook. This is explained in the
packagist site, will copy it here:

It is highly recommended to set up the _GitHub service hook_ for all your
packages. This reduces the load on our side, and ensures your package is
updated almost instantly. To do so you can go to your GitHub repository,
click the "Admin" button, then "Service Hooks". Pick "Packagist" in the
list, and add the API key you will find on your
profilehttp://packagist.org/profile/,
plus your Packagist username if it is not the same as on GitHub. Check the
"Active" box and submit the form.

Regards

Carlos

2012/8/8 raveren notifications@github.com

raveren:
I already did that earlier today:

http://i.imgur.com/OUtzo.png

still it does not acknowledge that I did :/

carlos-granados:
One question: when I output an object using your software, when I click on
the object name, the hierarchy is opened one level, but if I click on the +
sign the hierarchy is opened for all levels. Is this the expected behaviour
or am I doing something wrong? I find it strange that clicking on the plus
sign opens alls levels and it is not intuitive to click anywhere else to
just open one level... maybe you could add a configuration to choose this
behaviour?

Carlos

2012/8/8 raveren notifications@github.com

raveren:
Yeah, that is the intended and documented behavior, additionally if you doubleclick any plus symbol, _all_ collapsed elements on page will expand - probably hanging your browser in the process :)

Adding configuration options for this functionality would prove extremely difficult as the JS is not php-generated, but taken verbatim from a plain-text file. You could file an issue request and there's a chance I'd eventually get to it, but that's all I can promise unfortunately.

PR #15, opened 2012-08-12
Encoding selection support

deedweird:
Hi, Raveren! Thank you for such an amazing tool, debugging PHP code is so much better with Kint :)

The problems I ran into using Kint with Russian language (mostly due to PHP's broken mb_detect_encoding() ) and made a fix of:
1) With UTF-8 encoding Kint s() and d() showed incorrect string sizes of strings and d() output was garbage:
http://dl.dropbox.com/u/29937772/Kint/1_kint_raveren.png
Fixed: http://dl.dropbox.com/u/29937772/Kint/1_kint_me.png  

2) With WINDOWS-1251 encoding string lengths were reported correctly, but string previews for d() were broken:
http://dl.dropbox.com/u/29937772/Kint/2_kint_raveren.png
Fixed: http://dl.dropbox.com/u/29937772/Kint/2_kint_me.png

I should also note, that my patch makes a small regression: if the variables contain text in WINDOWS-1251, but Kint::$encoding is set to UTF-8(default), your version's s() and d() report correct strlen, while mine don't. Both versions truncate d() string previews.
Right strlen: http://dl.dropbox.com/u/29937772/Kint/3_kint_raveren.png
Wrong strlen: http://dl.dropbox.com/u/29937772/Kint/3_kint_me.png
I am not sure this can be fixed at all.


raveren:
Hey, deeweird, thanks for the pull request, I really appreciate both that _and_ the kind words.

I knew Kint was lacking in the character encoding department (as is my knowledge on the subject to be frank). 

I'll have to research this some more as I aim for the zero-setup goal and would like that Kint would work out of the box without setting up the encoding manually. Did I understand correctly, that using mb_\* functions with 'UTF-8' is _not_ the magic bullet?

And s() has been slightly left alone as I was planning to rewrite it to use the same parsing engine as d(), but in the meanwhile I really need it to help me develop the parsing engine itself. So bugfixes to s() are not yet a priority until I finish 1.0 final.

deedweird:
Oops, I accidentally sent an uncomplete message, sorry for that.

As far as I know, mb_\* functions work pretty well with UTF-8. But unfortunately, not all projects are in UTF-8. There is a lot of legacy (and/or language-specific) code, which uses WINDOWS-1251, KOI8-R, KOI8-U, CP866 and other old crappy encodings :( 

Also, the internal character encoding, which is get/set by mb_internal_encoding(), is quite a pain itself. AFAIK, it defaults to ISO-8859-1, aka latin1, which is a single-byte encoding, unlike UTF-8. So, this causes problems with non-latin characters and strings, when mb_\* functions are called without explicitly defined encoding. There are many projects, where internal character encoding is set to default.

I happen to work on such a project, it uses WINDOWS-1251(but internal encoding is set to default, which is ISO-8859-1), is big, complicated and full of turducken code. So I cannot just set internal encoding, as it may break something.

The ability to explicitly set the encoding is a saving grace in such cases, because mb_detect_encoding() is broken and cannot be used to find out the encoding of a non-UTF8 string.

raveren:
I think I solved this problem completely except of possibly string length mismatches. Changes are pending in the next build, thanks for your patience.

ma5ter:
This does not help with Cyrillic encodings, I've made a simple addition to override encoding using config file, this works fine enough.

Default mb_detect

Override in config


``` diff
--- R:\kint-master\config.default.php   2013-03-06 11:02:42.000000000 +0400
+++ R:\.kint\config.default.php 2013-03-14 09:10:30.000000000 +0400
@@ -120,6 +120,10 @@ $_kintSettings['theme'] = 'original';
 $_kintSettings['keyFilterCallback'] = null;


+/** @var string set to encoding you want all strings to be displayed */
+$_kintSettings['stringEncodingOverride'] = 'windows-1251';
+
+
 /** @var bool only set to true if you want to develop kint and know what you're doing */
 $_kintSettings['devel'] = false;
```

``` diff
--- R:\kint-master\Kint.class.php   2013-03-06 11:02:42.000000000 +0400
+++ R:\.kint\Kint.class.php 2013-03-14 08:56:08.000000000 +0400
@@ -26,6 +26,7 @@ class Kint
    public static $maxLevels;
    public static $enabled;
    public static $theme;
+   public static $stringEncodingOverride;
    public static $devel; # todo remove

    protected static $_firstRun = true;
```

``` diff
--- R:\kint-master\parsers\parser.class.php 2013-03-06 11:02:42.000000000 +0400
+++ R:\.kint\parsers\parser.class.php   2013-03-14 09:12:42.000000000 +0400
@@ -403,13 +403,17 @@ abstract class kintParser extends kintVa
    {
        $variableData->type = 'string';

+       if (!Kint::$stringEncodingOverride) {
        if ( function_exists( 'mb_detect_encoding' ) ) {
            $subtype = mb_detect_encoding( $variable );
            if ( $subtype !== 'ASCII' ) {
-
                $variableData->subtype = $subtype;
            }
        }
+       }
+       else {
+           $variableData->subtype = Kint::$stringEncodingOverride;
+       }

        $variableData->size = self::_strlen( $variable );
        $strippedString     = self::_stripWhitespace( $variable );
@@ -470,16 +474,16 @@ class kintVariableData
    protected static function _escape( $value )
    {
        # force invisible characters to have some sort of display
-       $value = preg_replace( '/[\x00-\x08\x0B\x0C\x0E-\x1F\x80-\x9F]/u', '�', $value );
+//     $value = preg_replace( '/[\x00-\x08\x0B\x0C\x0E-\x1F\x80-\x9F]/u', '�', $value );

        if ( function_exists( 'mb_encode_numericentity' ) ) {
            return mb_encode_numericentity(
-               htmlentities( $value, ENT_QUOTES, 'UTF-8' ),
+               htmlentities( $value, ENT_QUOTES, Kint::$stringEncodingOverride ? Kint::$stringEncodingOverride : 'UTF-8' ),
                array( 0x80, 0xffff, 0, 0xffff, ),
-               'UTF-8'
+               Kint::$stringEncodingOverride ? Kint::$stringEncodingOverride : 'UTF-8'
            );
        } else {
-           return htmlentities( $value, ENT_QUOTES, 'UTF-8' );
+           return htmlentities( $value, ENT_QUOTES, Kint::$stringEncodingOverride ? Kint::$stringEncodingOverride : 'UTF-8' );
        }
    }
```

raveren:
Do you get the Russian strings from external input or are your files encoded in windows-1251? I would like to get the source of the file that produces the empty output so I could experiment more on that.

raveren:
Ok, I think I got to the bottom of this, try pulling the latest version and adding 

``` php
$_kintSettings['charEncodings'] = array('windows-1251');
```

to your config.php

PR #22, opened 2012-09-14
BUG: splObjectStorage produced exception

carlos-granados:
splObjectStorage produced an exception. The problem is that the parser Factory method expects to receive a reference to a variable. We were calling it with the result of a function call and this cannot be used in a call by reference. Added a variable to call the function. Also removed an unused tmp variable


raveren:
awesome, thanks, I never had a real world occurrence of splObjectStorage myself..

PR #24, opened 2012-10-21
Added coloring and hyperlinks for recursive references

jthomaskerr:
Sorry about the delay, here is the mods that i proposed. I haven't included anything wrt to filtering, since you now have a filter callback which can do the same job.


jnvsor:
@jokester01au Welp. 4 and a half years late, but do you feel like updating your PR for the 2.x branch?

PR #32, opened 2012-12-05
Added function_exists('dd')

larrybolt:
Due to the framework (laravel) I'm using the function dd was already created, and even tho there is a check whenever the 'd'-function exists, there wasn't one for the 'dd' function as well.


raveren:
that's really unfortunate, but thanks

PR #43, opened 2013-01-19
Add a alias switcher.

menthol:
In somes cases we don't want kint to create the aliases. Now you can select aliases you want to.


raveren:
duplicate of #54

PR #49, opened 2013-02-19
Force a modifier for the next dump

ericthelin:
This is very useful so that you can build an alias that always assumes that a certain modifier is used. ie.

``` php
function dumpDeep()
{
    Kint::$modifier = '+';
    $args = func_get_args();
    return call_user_func_array( array( 'Kint', 'dump' ), $args);
}
```

### DIFF
--- a/Kint.class.php
+++ b/Kint.class.php
@@ -27,6 +27,7 @@ class Kint
 	public static $enabled;
 	public static $theme;
 	public static $devel;
+	public static $modifier = null;
 
 
 	protected static $_firstRun = true;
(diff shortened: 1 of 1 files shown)


raveren:
Thank you for the pull request, but I am in the progress of revamping the shorthand system which will allow you to rename `d`, `dd` and `s` functions, define new ones with arbitrary options/modifiers etc - all with simple configuration instructions. Unfortunately, this patch would interfere with the upcoming functionality, sorry...

ericthelin:
That is OK. Especially if you can accomplish my goal in the process.

That said you may be interested in adding what I am using the custom alias
for. I am forcing the @ modifier so that I can capture the output and add
it to a global array that I call kintDelayed (a static array would work
just as well). Then I register a shutdown function that does a foreach over
the array an prints it. The result is that all output is moved to the end
of the page where it doesn't interfere with the layout. I wonder if the
shutdown function could be replaced be a class destructor?

Eric

Btw, great job on the whole library! There are some really great tricks
that I never would have thought of. Especially adding the variable name and
origin of the dump. I have been wanting that for many years.

raveren:
That should be a nice feature to have indeed, I created issue #50 for this purpose, but I won't be implementing this too soon, there's lots of stuff to do still.

Thanks for the kudos, that's the main thing I do this for. I actually thought of a way of fetching variable names by parsing the source at the middle of the night and I got up to implement it. At that night Kint was born, I took Kohana's debug dump function and started a new class based on it. You can still see its remnants in `kintLite` - my most personally hated part of Kint source currently (mainly because it's out of place).

PR #52, opened 2013-02-21
Fixed buffer cleaning with nested buffers.

istvan-ujjmeszaros:


raveren:
I'm not fluent in `ob_*` functions, could you elaborate, what does this fix exactly and how is it better than the previous approach?

istvan-ujjmeszaros:
Basically nested buffer means that you start a buffer, and while the buffer is active, you start an other one. I am using it in hierarchical templates, so I can call a template from an other one and can manipulate the main template's content based on the sub template's requirements.

ob_clean() only clears the lastly created buffer, so the previously created buffer's contents are getting outputted before kint renders it's output.

The fix simply goes through all of the active buffers one-by-one, and drops all of their contents (ob_get_level() is maintained by PHP automatically, so there is no need to any extra action when using the fix).

From the [manual on ob_start()](http://php.net/manual/en/function.ob-start.php):
_Output buffers are stackable, that is, you may call ob_start() while another ob_start() is active. Just **make sure that you call ob_end_flush() the appropriate number of times**. If multiple output callback functions are active, output is being filtered sequentially through each of them in nesting order._

raveren:
Well that's embarrassing, thanks for the explanation, but when I glanced at your changes I thought it was the other way around - I somehow thought the lines you added were the ones you removed and vice versa and I was sure `while(ob_get_level())` was the correct way - thus the noob question. Sorry, and thanks again.

PR #55, opened 2013-03-16
Minor bugfix on dump() method

ulentini:
Hi, I started using your library and I find it quite useful.
While I was trying to modify it for my project I noticed a minor bug within `dump()` method in Kint class, so I fixed it.


raveren:
thanks!

raveren:
That's unfortunate, I must have miss-clicked and pressed cancel instead of merge previously, apologies.

ulentini:
No problem :)

PR #57, opened 2013-04-08
Fixed text align

vinculis:


raveren:
I've noticed this - issue comes up when dumping after some center-aligning html has been output. Thanks!

PR #60, opened 2013-04-11
Added classmethods parser

aljinovic:
I have added class methods parser, shows as additional tab for objects.

Example: http://www.opengluco.com/demo/kint/demo.php

It shows all class methods sorted by access type and name.

The "+" displays documentation block (if available for given method).

Return value is parsed from PHPDocumentor @return if available.

kint_classmethods

### DIFF
--- a/parsers/custom/classmethods.php
+++ b/parsers/custom/classmethods.php
@@ -2,18 +2,18 @@
 
 class Kint_Parsers_ClassMethods extends kintParser
 {
-	protected function _parse(&$variable)
-	{
-		if(!is_object($variable)) {
+    protected function _parse(&$variable)
+    {
+        if(!is_object($variable)) {
             return false;
         }
 
-		$reflection = new \ReflectionClass($variable);
+        $reflection = new \ReflectionClass($variable);
 
         $public = $private = $protected = array();
 
         // Class methods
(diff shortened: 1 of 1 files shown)


raveren:
Woah, this is awesome, thank you! 

I think it may make sense to cache the class definitions in a static variable, as they stay the same through out the request anyway and dumping traces - or objects in loops - would produce unnecessary overhead re-reflecting the class each time. On the other hand, as it is now, you can debug changing signatures of classes with Reflection. Not sure what's best here.

A definitely helpful addition would be to add links to method references: if they are inbuilt - to online PHP documentation - or `IDE://file:line` in case of custom ones.

PHP-REF https://github.com/digitalnature/php-ref has the "online documentation" feature in case you're interested.

Another minor thing, this does not display which methods are inherited, not even sure whether this would be useful. 

Again, this is all just nitpicking, top-notch code and functionality, thanks a lot!

aljinovic:
Thanks for merging it, I'm glad you like it!

I'll look into inheritance to see if it's doable and if performance would be the issue there maybe.

Static per-request caching is an awsome idea, that will improve performance with large object collections and nested objects.

links to file or documentation would be great, I'll look into it.

Thanks for the suggestions and for the awsome tool, it's been more thanhelpful. Keep up the good work!

PR #63, opened 2013-05-03
Fixed accidental misuse of $className variable in classmethods parser

aljinovic:
$className variable (line 13) is used as cache key, usage in method parameter parsing (line 36) caused fatal error and disabled caching.


raveren:
Goddamn it, thanks, no idea why it doesn't break when I use it, but it's a stupid bug, good catch!

aljinovic:
It breaks only when it was set to null (illegel offset) on some objects.

PR #65, opened 2013-05-24
Add a line break after s() output

will123195:
I just added a line break so `s()` output is easier to read.

Sample output with the line break:

```
<pre>integer 0</pre>
<pre>NULL</pre>
<pre>NULL</pre>
<pre>array(2) [
    'a' => string (1) "a"
    'b' => string (1) "b"
]</pre>
<pre>object stdClass (1) {
    public test -> string (1) "1"
}</pre>
```

Same output without the line break:

```
<pre>integer 0</pre><pre>NULL</pre><pre>NULL</pre><pre>array(2) [
    'a' => string (1) "a"
    'b' => string (1) "b"
]</pre><pre>object stdClass (1) {
    public test -> string (1) "1"
}</pre>
```


raveren:
I've actually already rewritten s() and sd() to use Kint parsers and decorators, so it accepts modifiers and has other cool stuff. I just need to test it more. I've merged your request regardless.

PR #66, opened 2013-05-30
Tweaked Solarized themes

Rarst:
I have noticed Solarized themes colors are slightly off (especially light one being uncharacteristically dark) and reshuffled them to standard pairs.

Plus small bugfix for code blocks, rule was losing to external stylesheet and rendered text unreadable.


raveren:
ah much appreciated, I've no sense of aesthetics at all

Rarst:
Not that I do, that's why Solarized has handy [implementation guides](http://ethanschoonover.com/solarized#usage-development) on which color goes where for people like us. :)

PR #71, opened 2013-06-06
FIX: Crash on @thobbs/phpcassa objects with UUIDs

mcd-php:
Error was:

```
PHP warning
mb_strlen(): Unknown encoding ""
raveren/kint/parsers/parser.class.php(559)
```


raveren:
somehow missed this one, sorry for taking so long.

PR #72, opened 2013-06-07
added support for displaying objects' __toString() methods

winkbrace:
by adding
- config displayToString (default set to true for this fork)
- string display in object header row in parser.class.php
- custom parser to display full toString as a seperate tab
- test for displaying


jnvsor:
Nearly 4 years late, but would you like to implement this as a plugin in Kint 2.x?

winkbrace:
No, sorry. I don't use kint anymore. ;)

Issue #46, opened 2013-02-12
Ability to drag out kint output to a pop up window for future reference and whatnot


raveren:
``` javascript
window.open(
    'data:text/html;base64,'
    + window.btoa(
        unescape(
            encodeURIComponent(
                document.documentElement.innerHTML
            )
        )
    )
);
```

Instead of `document.documentElement` use needed Kint nodes along with something along the lines of `document.querySelectorAll('script')[0].innerHTML`.

aljinovic:
I implemented this feature in my fork. Base64 was a bad idea because it breaks browser when there's a lot of debug data, so I implemented it this way:

``` javascript
newWindow = window.open();
if(newWindow) {
    newWindow.document.open();
    newWindow.document.write(html);
    newWindow.document.close();
}

```

I fixed some compatibility issues with older PHP versions that don't support "static::" and "$var::" static calls.

Also, I fixed the bug with `_stepIsInternal()` method: didn't detect `call_user_func_array` inside Kint class as internal when short functions were called.

Result: http://www.opengluco.com/demo/kint/demo.php

Raveren, I'll make a pull request in 1.0.0-wip branch if you think this is OK.

Could you just tell me which minifier script you're using for .js so I can minify it again before merge since there are some changes inside also. Thanks!

raveren:
Wow, dude, that is terrific! Sure, please do a pull request, these changes are totally awesome.

I use a custom File Watchers Php storm plugin config, parameters of which are at the top of _kint.js (I know this sucks, I need to brush up contribution guidelines heavily).

You can however just use the [online compiler](http://closure-compiler.appspot.com/home) on advanced settings and then wrap the output inside `(function(){%output%})()` - the online tool does not support this configuration option.

raveren:
Even the dumped object in your demo is awesomely crafted to showcase most of Kint features, that's just wonderful :)

aljinovic:
I was planing of making an interactive demo where you can switch themes and see the actual code behind, if you agree ;)

raveren:
That would be stellar! :)

aljinovic:
I've created a simple demo with theme switcher: http://www.opengluco.com/demo/kint/demo.php

Any suggestions?

raveren:
Wow you even created a new theme :D Fucking A!

 I'm finishing up with my own batch of changes and will comment further tomorrow.

aljinovic:
My colleague did the theme, he'll make a pull request when it's done.

[...]

PR #86, opened 2013-10-22
implemented "open in new window"

aljinovic:
#46
- implemented "open in new window"
- fixed the bug with `_stepIsInternal()` method which didn't detect call_user_func_array inside `Kint` class as internal when short functions were called
- fixed some compatibility issues with older PHP versions that don't support "static::" and "$var::" static calls


raveren:
I'd be really thankful if you could explain a reproduceable case of 

> fixed the bug with _stepIsInternal() method which didn't detect call_user_func_array inside Kint class as internal when short functions were called

aljinovic:
When you call dd($var) from some_method, output should be:
"Called from /mnt/cifs/kint_master/demo.php line 69 in (some_method())"
Before, it was:
"Called from /mnt/cifs/kint_master/demo.php line 69 in (call_user_func_array())"

Reason was this line in _stepIsInternal():

``` php
return in_array( strtolower( $step['function'] ), self::$aliases['functions'], true );
```

All d(), dd(), ... methods contain "call_user_func_array()", and that method gets called from inside the class, but doesn't belong to a class so it's not detected as internal.

I added the check:

``` php
$fromThisFile = isset( $step['file'] ) && $step['file'] == __FILE__;
```

this just checks if the call was made from inside the kint class file and flags it as internal.

raveren:
Aw fuck, I've already fixed that previously in another way, but did not commit the changes strangely :/

raveren:
ah, also, I couldn't make your auto fetching of title work with traces - and it had some other very minor issues, so I substituted it for current date time for now.

PR #88, opened 2013-10-24
added kint overview example page

aljinovic:
- added kint overview example page
- replaced static:: with self:: - php 5.0 doesn't support it

currently available here (would be cool to have it on main page): http://transfercroatia.com/transport/web/kint/examples/overview.php

### DIFF
--- a/Kint.class.php
+++ b/Kint.class.php
@@ -554,14 +554,14 @@ class Kint
 	private static function _stepIsInternal( $step )
 	{
 		if ( isset( $step['class'] ) ) {
-			foreach ( static::$aliases['methods'] as $alias ) {
+			foreach ( self::$aliases['methods'] as $alias ) {
 				if ( $alias[0] === strtolower( $step['class'] ) && $alias[1] === strtolower( $step['function'] ) ) {
 					return true;
 				}
 			}
 			return false;
 		} else {
-			return in_array( strtolower( $step['function'] ), static::$aliases['functions'], true );
+			return in_array( strtolower( $step['function'] ), self::$aliases['functions'], true );
 		}
 	}
 
--- a/examples/overview.php
+++ b/examples/overview.php
@@ -1,0 +1,175 @@
+<html>
+<?php
+require('../Kint.class.php');
+
+$selectedTheme = isset($_GET['theme']) ? $_GET['theme'] : 'original';
+$allowedThemes = array();
+$dh  = opendir('../view/inc');
+while(($filename = readdir($dh)) !== false) {
+     if(strpos($filename, '.css') !== false) {
+         $allowedThemes[] = str_replace('.css', '', $filename);
+     }
+}
+
+if(in_array($selectedTheme, $allowedThemes)) {
+    Kint::$theme = $selectedTheme;
+}
+
+class BaseUser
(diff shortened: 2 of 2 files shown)


aljinovic:
There's a little compatibility issue with older versions of PHP that don't support namespaces.

In new code, some instances are created with backslash, ex: `new \ReflectionObject`, parser.class.php, line 364.

My old server has old PHP and Kint stopped working. I don't know if you're planing on using namespaces in Kint and loose support for older PHP versions, but it could make a problem for some people stuck on some legacy code and servers that can't upgrade that easily.

Also, method http://php.net/manual/en/reflectionproperty.setaccessible.php was introduced in PHP version 5.3, so it will cause problems too.

Errors: http://www.opengluco.com/kint/examples/overview.php

raveren:
Wow dude, you're great! Yeah I'm definitely keeping compatibility **as wide as possible**, I've really put a lot of effort into it.

The `static` keyword is an interesting issue, it might prevent kint from being extendable, but I'll sort it out in the future..

aljinovic:
I've been busy ;)
Here's a new theme: http://transfercroatia.com/transport/web/kint/examples/overview.php?theme=gray

raveren:
Wow, great job! I've stayed up late yesterday too ;) but I'm not ready to release my changes..

The theme is excellent, there are some spacing issues, which need a little polishing, but it looks really good. The black one lacks a _little_ contrast IMHO to be readable enough, but it's up to you and I'll accept the pull requests as you think they should be :)

Thanks again, I'll credit you wherever possible!

aljinovic:
I didn't do the black one. I can fix the spacing, just provide a screenshot so I can see what you're talking about ;)

raveren:
here you go:

http://i.imgur.com/MLlj9r7.png

aljinovic:
Nice catch with that spacing, check if it's ok now, same link.

I have a question about small "open in new window" arrow, was it intended only for trace? What should click on it do in normal debug (open only small trace in new window or should it be removed from there)?

arror

raveren:
Well the small arrow is always there because you will sometimes want to pop-out outputs that are not collapsed - and it also gives a different result: the callee line is left present. I agree that it looks more cluttered, I'll have to ask my designer friend to come up with something.

Good job on the spacing, but here are other miniscule details I've noticed:

http://i.imgur.com/S1mTnSH.png

No 1. is present in two places (I think) and the empty cell in tabular data has the default background.

aljinovic:
I think I managed to fix the problem with positioning, same link.

raveren:
Hm, the tabs still look wonky.

aljinovic:
What do you mean exactly?

raveren:
Sorry for my brevity, I'm at work now. 

I must have looked at an older version, it looks fine now, trace tabs are still off a little though: 

http://i.imgur.com/HhETRLO.png

raveren:
I mean the leftmost border

aljinovic:
OK, I think I fixed it now ;) Didn't understand the problem at first...

http://transfercroatia.com/transport/web/kint/examples/overview.php?theme=gray

If you agree, I'll make a pull request for it.

PR #89, opened 2013-10-28
gray theme

aljinovic:
- new theme (gray)
- small changes in overview example

Example: http://transfercroatia.com/transport/web/kint/examples/overview.php?theme=gray


raveren:
Could I ask for a less-generic name? Something like aljinovic-grey? For future-proofing :)

aljinovic:
Could it be some name that doesn't look like ego-trip ? :)

raveren:
It's totally up to you :) Get creative or `Ante-is-awesome-grey`, no difference for me :))

PR #92, opened 2013-10-30
changed tabs layout in aante-light theme

aljinovic:
Demo: http://transfercroatia.com/transport/web/kint/examples/overview.php?theme=aante-light


raveren:
pushed some changes myself, did you know about `microtime()`-based profiling? 



I must make a pretty presentation of all Kint's features someday.

aljinovic:
cool, that's a nice feature, you should add it to demo also, it's a nice feature to have, no need to worry about scope of first microtime ;)

raveren:
Yeah, it takes care of it, only shows average duration on 3rd call for example. 

Since some parts of the application at my job deal with Big Data, I use this feature a lot - to locate bottlenecks when things get slow :)

aljinovic:
same here, but I never thought of implementing it into kint, great idea!

raveren:
Thanks, it's really easy to use too:

``` php
d(
    microtime(),
    the_method_you_are_profiling(),
    microtime()
);
```

PR #93, opened 2013-11-04
Last callee fix in _getPassedNames()

alexander-danilenko:
If we going thru whole backtrace we need to set previous passed element as callee instead just breaking cycle and save current callee as needed callee. 

It fixed problem in my Drupal module:
"Called From" string contains last called function place (kint() - custom alias) instead current place when i execute my kint() function.

Example of alias, calling of one will prints kint::dump() without information about called from place:
https://gist.github.com/fat763/7306086

Please, comment this situation.

Thank you!

### DIFF
--- a/Kint.class.php
+++ b/Kint.class.php
@@ -392,6 +392,7 @@ class Kint
 				( isset( $callee['class'] ) && strtolower( $callee['class'] ) === strtolower( __CLASS__ )
 					&& strtolower( $callee['function'] ) === 'dump' )
 			) {
+				$callee = $previousCaller;
 				break;
 			} else {
 				$previousCaller = $callee;


raveren:
Hi there, I'm really sorry, but as [stated in the readme](https://github.com/raveren/kint#contributing), I prefer to accept merges to and from the [wip branch](https://github.com/raveren/kint/tree/1.0.0-wip) only. 

I think this fix is not only already there, but I already rejected a similar pull request.

I'm planning to publish the branch to stable status in the course of a month. It has a plethora of new features, you should check it out. For example in `$kint_markup = @Kint::dump($args);` you no longer need the `@`, Kint automatically detects that it's being assinged to a variable and returns the output instead of dumping - the `@` is optional, but that's just the tip of the iceberg.

alexander-danilenko:
Oops, my apologies. I'll check this out. Thank you!

PR #112, opened 2014-04-06
Added 1.0 Branch Alias To `1.0.0-wip` Branch

GrahamCampbell:
Also, is there any news on when this will be released and merged into the master?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,5 +20,10 @@
     },
     "autoload": {
         "files": ["Kint.class.php"]
+    },
+    "extra": {
+        "branch-alias": {
+            "dev-1.0.0-wip": "1.0-dev"
+        }
     }
 }


raveren:
Hey, not familliar with composer, what impact does your change make?

Also, I'm feeling terrible as it is, I'm swamped with real life and am trying to put in as much free time as possible into Kint, but there's been just too little of it, I'm getting there, please be patient...

GrahamCampbell:
Sure, I'm not trying to rush you. My change means we can use `1.x` or `1.0.x` or `~1.0`, or any other valid version constraint to require the 1.0 version provided minimum stability is set to dev, or we can just add `@dev` to the requirement so `1.x@dev` would work too rather than trying to require the branch. This means that when you do release 1.0, nobody need make any changes to their composer.json files because they were already requiring the 1.0 branch. You can see an example of this here: https://github.com/panique/php-login/pull/405/files.

raveren:
Uh-huh, I understand some of these words :D

But seriously, composer seems like an excellent tool, I'll try it out on my next freelance project.

GrahamCampbell:
:)

PR #119, opened 2014-06-10
1.0.0 wip

### LINKED ISSUE #118: Disable JS and CSS output when not needed
I have Kint included in my php.ini file for all scripts run on my dev server.  Because of this, Kint javascript is output at the bottom of AJAX and other text-only output which will break the script receiving that output, even if there were no traces or debugs in the script.

I understand that AJAX sniffing is now an option, but it would be nice if Kint only output the relevant JS and CSS when it's needed because a debug or a trace was actually called.

benjamw:
The main point of the pull request is the fix to stop outputting the scripts to files when they aren't needed.

Fixes #118

### DIFF
--- a/Kint.class.php
+++ b/Kint.class.php
@@ -79,6 +79,8 @@ class Kint
 		)
 	);
 
+	private static $called = false;
+
 	protected static $_firstRun = true;
 	protected static $_headerNo = 0;
 
@@ -266,6 +268,8 @@ class Kint
 			}
 			return '';
 		}
+
+		Kint::$called = true;
 
 		echo $output;
 		return '';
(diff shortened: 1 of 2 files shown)


raveren:
as I mentioned in the other issue, you made an easy-to-miss, but fundamental misunderstanding of the functionality, and it's disabled by default in the future. Thank you anyway and sorry for wasted effort :(

benjamw:
It wasn't wasted. I use it on my server.

I don't really use the AJAXable functionality that you describe in #118 because any debugging I would do for scripts that return AJAX, I would do directly IN the AJAX response. So it works for me.

Thanks for an awesome script.

FYI, there are other changes not related to the direct issue at hand in those commits. I would have submitted separate pull requests for each, but couldn't figure out how to separate them in Github.

raveren:
Yeah, I had a similar problem with github merging multiple commits into one pull request.

Thanks for the overview fix, I'll implement it manually!

The ajax feature is precicely for debugging ajax, you can click on the link to resend the same request after making changes and see the new output, but frankly even I don't use it that much - I open the request in a separate window :) 

Although it's helpful when the backend expects the request to be ajax and differs the output if it is not...

This feature will be someday helpful when Kint can be used as an error handler, but that's not so soon as you expect :)

PR #122, opened 2014-07-10
Allow users to specify which IPs are allow to debug

aram535:
A simple check to see if the requested IP address is in the array.  This allows us to point the config variable into a database or load it via a static list of develop ip addresses.  In our code we use a local Kint::enable to allow debugging or not.  Works in development and UAT.

### DIFF
--- a/Kint.class.php
+++ b/Kint.class.php
@@ -34,7 +34,6 @@ class Kint
 
 	# non-standard function calls
 	protected static $_statements = array( 'include', 'include_once', 'require', 'require_once' );
-
 	/**
 	 * getter/setter for the enabled parameter, called at the beginning of every public function as getter, also
 	 * initializes the settings if te first time it's run.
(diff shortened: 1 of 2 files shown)


mateuszjarzewski:
I think that authorization should be handled elsewhere.
For example code below will do it just fine.

``` php
$username = $_SERVER['PHP_AUTH_USER'];
$password = $_SERVER['PHP_AUTH_PW'];
if ( 'admin' === $username ){
    if ( 'password' === $password ){
        Kint::dump($_SERVER);
    }
}
```

IMO it's nonsense to put this kind of logic into Kint project

aram535:
This is actually the result of a consensus of about 25 people in our local PHP group, whom all use Kint and they all use some version of this patch.  You want to be able to put any page into debug via a small configuration file or database.   The local solution is then wrap each page's debug on or off based on global debug variable as well as checking locally for a _GET param.  So it works great for all solutions.  If you don't want to specify an IP you can do as said above, if you want to use it, it's available.

mateuszjarzewski:
They all wrong.. This should be handled by application authorization object or simple block of glue code like I pasted above.

raveren:
Hi there folks, I have to go with @mateuszjarzewski here, Kint only provides the `enabled()` method to use at own discretion. It should go into the application bootstrap process as the use cases are unlimited and I leave this to the developer to decide without me favoring one approach over the others.

I personally don't solely use the IP based approach, but a combination of state of obscure cookies, host address and/or the user IP based on the security requirements of the application. 

The whole ordeal, BTW, is used to also toggle displayed errors and other application debugging information. I basically turn on Kint when the app is running in development mode, but that is just one of the ways to tackle this issue.

PR #125, opened 2014-07-27
add clean style to css and add ability to dump to file.

liverbool:


liverbool:
Warning! I've changed composer.json

raveren:
Hey, thanks a lot for your hard work, but there are some problems with the pull request.

First of all, you forked the `master` branch instead of `1.0.0-wip` - contrary to what I specifically ask in the readme.

Secondly, you commited a reformatted file so I cannot see which are actual changes made by you - and which are just altered whitespace.

Third, the `$tofile` parameter goes against the ideology of Kint and can produce terrible, unexpected outcomes for users who pass a file name to dump - it would be overwritten without the explicit intention of the user.

Thank you for the theme, but I cannot use it either, as themes were rewritten from the ground up in the `wip` branch.

I'm going to have to reject this pull request, I'm really sorry.

PR #126, opened 2014-08-04
Update parser.class.php

### LINKED ISSUE #117: Solution for cannot redeclare class kintParser
Im using v 0.9.. and i found this error. "cannot redeclare class kintParser in kint/parsers/parser.class.php".. the problem is that kintParser trying to extend kintVariableData which is not yet declared.. the soution is simply to swap the position of class of kintParser and kintVariableData which are both on the same file; parser.class.php. Sorry i dont know how to make a pull/fix request.

efalder413:
swapped class order to fix #117

### DIFF
--- a/parsers/parser.class.php
+++ b/parsers/parser.class.php
@@ -1,4 +1,127 @@
 <?php
+
+class kintVariableData
+{
+	/** @var string */
+	public $type;
+	/** @var string */
+	public $access;
+	/** @var string */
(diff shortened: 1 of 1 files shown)


raveren:
Alright, thanks a lot! Which PHP version are you using by the way, I never encountered this bug?

efalder413:
5.5.9-1ubuntu4.3

PR #130, opened 2014-09-04
Switch from Composer Files autoload to classmap

jcisio:
Using the "Files" classloader always includes the file in every request. We should use the more appropriate Classmap loader.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "php": ">=5.2.0"
     },
     "autoload": {
-        "files": ["Kint.class.php"]
+        "classmap": ["Kint.class.php"]
     },
     "extra": {
         "branch-alias": {


and1truong:
:+1:

jcisio:
It only breaks if you use a lazy class loader _and_ dd. If you just do @require '/kint/Kint.class.php';@ then there's no problem.

I think it's useless to use class loader with a function in global scope. If we insist to do that, we can split function declarations in a separe file and people want to use shorthand functions can include that file.

jcisio:
A clear separation could also fix issue like https://github.com/raveren/kint/issues/110 without making a "fake" class file ;)

raveren:
I get what you're saying, but 

> people want to use shorthand functions can include that file.

the ideology of Kint is to make it zero-setup. All core features should work flawlessly without any tinkering from the user. The side effect of including a file (as it is now) that does not impact anything on every request is well within reason.

jcisio:
Module like this https://www.drupal.org/project/devel does:

```
/**
 * An alias for drupal_debug().
 */
function dd($data, $label = NULL) {
  return drupal_debug($data, $label);
}
```

and does not check if `dd` is already defined is not compatible with Kint, unless Kint is loaded after devel. And it is impossible to do that with a class loader (which runs very first, and loads Kint at the beginning).

Of course I can ask that module to an `if function_exists()`, but in this case we get undeterministic function dd that depends on loading order.

The idea of Composer is to namespace everything. So I think if we want to support Composer, then we should not add functions in the global scope, at all.

raveren:
I was actually in the talks with the developers to include Kint into Drupal 8.0 Devel. But I need to come up with a different Kint configuration and extending model first...

> The idea of Composer is to namespace everything. So I think if we want to support Composer, then we should not add functions in the global scope, at all.

Removing `dd` is out of the question, but I also added a `ddd` alias to the same functionality, I think it should be in Github already. Anyway this is a debugging tool - it does not have to follow all the same guidelines that apply to other - traditional - tools. It's a meta-project - used to create tools and the first and foremost priority is ease of use in everyday tasks.

jcisio:
Well, I didn't mean to remove dd. I mean 1/ either to move it (and friends) into a separate file 2/ or keep it as is and does not force Composer users to use it.

The 2/ choice is what this PR does. If Composer users want to use those shortcuts, then a simple `require` as written in Kint's README will work.

raveren:
Serious question, what would be a _real wold_ benefit or problem solved by doing this?

jcisio:
Actually, it causes a fatal error with redeclared function when using Kint with Composer + Devel module in Drupal 6 or 7. And I think it is a normal setup:
- I have a Drupal website with Composer and it class loader. I fetch Kint with Composer.
- With the Files class loader, Kint is loaded and dd function is declared.
- Then Drupal bootstraps its modules, included Devel.module that defines the dd function again. And bang!

I now have to patch the class loader to fix this incompatibility.

jcisio:
In a less common setup, Drupal user want to include Kint.class.php directly somewhere (in settings.php). The same bug happens. To fix it, without class loader, they have to include Kint in a hook_init() so it is included _after_ devel.module. But it makes Kint unavaible in an earlier stage.

raveren:
I can definitely see the problem here, thanks. I was never aware Kint does not work with Drupal 6&7.

However, there's another segment of users who use Kint exclusively via `dd` and when this change goes live, they all will start receiving FATAL errors. In this scenario, I'd rather keep things as is to not screw existing users vs catering to new ones.

I can see one alternative solution - add a notice text to `dd` output that the function is going to be deprecated in favor of `ddd` (which hopefully is not used as widely..?) - and after some months remove the function entirely - stating that in the documentation too.

What are your thoughts?

jcisio:
Ah I see, a lot of people uses Kint with Composer https://packagist.org/packages/raveren/kint. Then I suggest this went into the next 1.0.x versions instead of 0.9. And we are lucky that we are near 1.0 ;)

It's because I don't know if ddd will help it avoiding name clash better. Why just disable all shortcut functions by defaut, and add a KintHelper class (which can be also autoload using Composer autoloader, and can also load the main Kint class for those who don't have autoloader). This class has a static method KintHelper::createShortcuts() that defines thoses function. So people who want thoses shortcut just call the method at the very beginning of their script.

raveren:
Again, ease of use and zero-setup are my main goals in Kint - I disregard performance considerations, naming conventions, allow myself to hack into the easily breakable PHP engine, output non-valid HTML, etc - all for those two goals (there are a few more, but the latter two are top priority).

I'm sorry, but I will not abandon the shortcuts, they are one of the things that make Kint effortless to use.

I will, however, most probably deprecate `dd` with a warning in the 1.0 version and remove it entirely in 1.1...

Thanks a lot for your constructive input, you sure know your stuff, and none of your arguments I disagreed with. However I have valid arguments for my implementation too, and being the author, I will choose what seems right to me, I hope you understand.

PR #142, opened 2015-04-10
Added log to file

hgati:

PR #146, opened 2015-06-05
introduced l() function for logging

andrejsstepanovs:
I saw that this function is still missing in kint repository.
As we discussed, logging functionality is really useful.
I am using this code for view weeks now and it is working flawlessly.

### DIFF
--- a/Kint.class.php
+++ b/Kint.class.php
@@ -830,3 +830,38 @@ if ( !function_exists( 'sd' ) ) {
 		die;
 	}
 }
+
+if ( !function_exists( 'l' ) ) {
+	/**
+	 * Logs variable information to file; accepts arbitrary number of parameters
+	 *
+	 * [!!!] this function uses undocummented Kint configuration settings that
+	 *       may become deprecated with future updates
+	 */
+	function l()
+	{
+		if ( !Kint::enabled() ) return;
+
+		if ( !in_array( 'l', Kint::$aliases['functions'] ) ) {
+			Kint::$aliases['functions'][] = 'l';
+		}
+
+		// adjust Kint settings manually, so this function accepts arbitrary number of
+		// parameters as we can't use call_user_func_array with Kint's inline modifiers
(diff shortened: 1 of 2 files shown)


raveren:
Thanks for the pull request and although the code works and suits your needs, it is sadly not of production quality - mainly not until the location of the log can be changed, and there's no elegant way to do that - yet. I'll be working on Kint configuration in the coming months..

I don't want to critique the code, it's fine, but to get a taste of what would make it production ready, these are just some of the potential problems I can see at a glance:
- uses short array syntax and Kint strives to keep PHP5.1+ compatibility.
- uses date function when it can cause PHP Warnings when timezone is not set in `php.ini`
- does not check permissions to write log file - PHP Warnings again.

sukrosono:
Hi guys,

I also see #142, i am just on same interest to see logging functionality on kint.

Here is my thought about log location and it's config. It's already nice that it aim os temp directory as it's default log loc.

@raveren I think basic way is good enough (kint provide conf file), while i am more concern about designing an elegant way to logging that works both for develop and production mode.

IMO, prob number:
2. That's the way php works, it's kint user responsibility to change and adjust it.
3. Basically it's same as number 2, but we want to be more friendly to people who just learning programming. Also, this one is affecting the main feature of logging itself. So we handle it with a proper error message (if necessary) if am fine with the default error message, and a couple of paragraph on documentation or readme about file permission.

I am waiting for your response guys, i think i have some hours free on this week.
Let me know if i can do some work to speed up this one to be merged. :blush:

jnvsor:
@brutalcrozt There have been several major rewrites in the time since this, and if you asked me nowadays I'd tell you to make a logging renderer instead of putting it in a helper function.

I'd also say this should be a separate repo with logging renderers for pdo/file/syslog/monolog/etc.

In any case, if you want to start working on this from scratch, feel free to throw in a PR and I'll try to give advice

PR #150, opened 2015-06-15
Remove self::_char( "═", 80 ) and Title if $returnOutput = true

rubao:
\Kint::$displayCalledFrom = false;
\Kint::enabled( \Kint::MODE_WHITESPACE );
\Kint::$returnOutput = true;
$output = \Kint::dump( $data );

but I still have the ======== x80 in the end and [ $varName] in the begin.

### DIFF
--- a/decorators/plain.php
+++ b/decorators/plain.php
@@ -223,6 +223,7 @@ class Kint_Decorators_Plain
 
 	private static function _title( $text )
 	{
+		if ( Kint::$returnOutput ) return '';
 		$escaped          = kintParser::escape( $text );
 		$lengthDifference = strlen( $escaped ) - strlen( $text );
 		return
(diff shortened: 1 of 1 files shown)


raveren:
Hey there, it's a nice idea, but this should be a separate option, as it changes existing functionality some users may already rely on. Configuration rework is coming to Kint, and I'll add a new option based on this suggestion when I'm done.

Sorry I can't just merge this in right away...

rubao:
hi raveren,
you are right, should be a separate option.
when you are planing to release the new version?

rubao:
hi @raveren maybe I can add this new option, what you think about `$returnTitle = true //default` and if someone need `$returnTitle = false`

jnvsor:
You can remove decorations as of aaba58d5eb397e4ee0a286c322138c76ed0e3933 (2.0-alpha2)

The title still appears as the first line of the output. If you still want to get rid of the title you can write a renderer extending `Kint_Renderer_Text`, stub out the `renderTitle` method, and replace the text renderer in `Kint::$renderers[Kint::MODE_TEXT]`. Or you could just write a wrapper function that strips the first line off :)

PR #153, opened 2015-06-25
Fixed detecting of string encoding based on Kint::$charEncoding setting

userlond:
New version ignores $charEncoding setting. I fixed it pasted necessary code from 0.9v branch.

### DIFF
--- a/parsers/parser.class.php
+++ b/parsers/parser.class.php
@@ -632,44 +632,29 @@ class kintVariableData
 	 * HELPERS
 	 */
 
-	private static $_supportedCharsets = array(
-		'UTF-8',
-		'Windows-1252', # Western; includes iso-8859-1
-		'euc-jp',       # Japanese
+    protected static function _detectEncoding( $value )
+    {
+        if ( function_exists( 'mb_detect_encoding' ) ) {
+            $mbDetected = mb_detect_encoding( $value );
+            if ( $mbDetected === 'ASCII' ) {
+                return 'ASCII';
+            }
+        }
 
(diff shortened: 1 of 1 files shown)


raveren:
ok, disregard most of my previous comment, I did not see you made a pull request, I commented on a commit I saw you make and didn't even see the new PR (wtf, github).

Anyway, the `$charEncodings` should have safe defaults that are now `$_supportedCharsets` if the user does not provide any. Did you use this functionality, was it useful?

userlond:
Hi, thanks for reply.
I'm working with some old-plain cyrillic projects :) Their default encoding is windows-1251. Then I use Kint with default settings, it outputs windows-1251 strings with non-latin symbols (like матрешка) as unreadable artifacts.

To fix this, I included Kint like this:

``` php
    <?php
    // include composer autoload
    require_once $_SERVER['DOCUMENT_ROOT'] . '/classes/vendor/autoload.php';
    // suggest variants of string encodings to kint
    Kint::$charEncodings = array('windows-1251', 'utf8');
```

With supplied settings, Kint outputs windows-1251 strings correctly and falls back to utf8 if needed.

In new branch setting was deleted, so I'v manually returned this function to one of projects and it works now.

About pulls. It's my first pull on github, so I may probably made smth wrong :) 

With first commit I removed $_supportedCharsets settings and returned _detectEncoding function from 0.x branch. So my code snippet above works fine.
With second commit I replaced space indents with tabs.

Hope this info helps.

About usefullness of feature "setting encoding variants manually". I think that autodetecting of encoding is too hard problem. And setting encoding is the great feature for some non-utf8 projects.

P.S. Thank for Kint, it's great.

PR #155, opened 2015-07-03
Added other way how to install via composer

royopa:


raveren:
This is already in the readme, but thanks :)

PR #156, opened 2015-07-03
Check for method_exists getClosureThis

eimajenthat:
Looks like that function was added in 5.4: http://php.net/manual/en/reflectionfunctionabstract.getclosurethis.php

My PHP 5.3 environment (we'll upgrade soon, but we haven't gotten to it yet) chokes on this call.  This check fixes the problem for me.  Packagist says you support back to 5.1.  If you only claimed support for 5.4 and up, I wouldn't request the change.  Thanks for the excellent library!


raveren:
thank you! :)

PR #174, opened 2015-11-13
Suggestion for code PSR Standard :)

karlpatrickespiritu:


jnvsor:
The wordpress-ey coding style drowning in whitespace is far from aesthetically pleasing to me, but it should be up to @raveren since he's likely to be the one doing the most work on it.

That said: :+1:

raveren:
@jnvsor I've never worked with WP, so the code style is the result of years of fine-tuning and finding the most readable format for me.

That said, I reject the thoughtful suggestion.

karlpatrickespiritu:
yeah, some frameworks doesn't even implement strict PSR standards. well, I just want to say that this library really helped me a lot on my projects. thanks @raveren :)

PR #176, opened 2015-11-17
Move _firstRun into decorators to ensure styles are output

jnvsor:
Keeps styles outputting once per decorator, but ensures they are output at least once each.

Before, if you called `s()` first the styles and javascript for `d()` wouldn't load, now they both load once.

### DIFF
--- a/Kint.class.php
+++ b/Kint.class.php
@@ -72,8 +72,6 @@ class Kint
 		)
 	);
 
-	private static $_firstRun = true;
-
 	/**
 	 * Enables or disables Kint, can globally enforce the rendering mode. If called without parameters, returns the
 	 * current mode.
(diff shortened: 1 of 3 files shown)


raveren:
Hey, thanks **a lot** for the work, before I do a code review, does this change take into consideration multiple `d` outputs in the same page, JS should never be output twice, it would break event listeners and whatnot...

jnvsor:
It does, that's the whole point of `firstRun` after all :)

This just moves firstRun into the decorators so that `init` is run at least once per decorator but not twice

Before:

```
s();
d();
s();
d();
```

Would give you:

```
<s init><s><d><s><d>
```

Now it gives you

```
<s init><s><d init><d><s><d>
```

raveren:
I'm merging this into my own local copy and will organically use it for a while - most probably a month or more. I use Kint in a lot of various contexts - logging, error reporting and not only debugging. If all continues to work as it is now (or better - and I _do_ encounter d init missing in very specific cases, so it's a real issue), I will press the tempting merge pull request button :)

Thank you, this is actually a difficult problem I tried to fix multiple times and did not invest effort since it works most of the time(tm) now :))

jnvsor:
If you're interested, I'll leave a [another branch](https://github.com/jnvsor/kint/tree/decorate-js) here. I was planning on making it a PR after this was merged, but I'll probably forget if it's a month from now :)

raveren:
Oh wow, you're awesome! Thanks for this, the feature looks handy - I'll try it out, and if it works as advised, to which I have little doubt, I'll ask for you to make another PR, for the proper attribution.

Fun fact, I toyed around with background Kint dump display - if Kint detects it's called in an ajax context, it would return its output as series of response headers (not 'echo') - and meanwhile every "normal" page has JS code injected to handle that. It resulted in a modular window (think "[colorbox](http://www.jacklmoore.com/colorbox/)") containing all dumps produced by Kint that were thrown in the background ajax call.

It worked very well, was a fairly unique and useful feature, but I had to rely on users to always set headers properly (to denote ajax calls) and generally not fuck up :) Which, as I realised with my tiny test group was way too much to ask and I removed the functionality. It proved to be too much hassle to be useful to everyone but me :) 

In the end, it's impossible to execute properly - short of writing a browser extension (to avoid the JS request handler code at the end of every page that Kint detects as not ajax), but FirePHP attempted to do that and has since been abandoned.

The code for this functionality is still residing in the history of some repository somewhere...

jnvsor:
Note: This requires 5.3 so might have to be delayed to a new major version (But really, 5.3 was 10 years ago...)

Or it could be rewritten into a static method but honestly thinking long term I'd rather ditch 5.1/2 instead

jnvsor:

PR #177, opened 2015-12-24
Display access path along with publically accessible variables

### LINKED ISSUE #175: Show access path on double click
This is a feature in a modified version of Krumo in Drupal's Devel module and I'm still surprised I don't see it anywhere else since it's so handy: Show the access path below the variable on double click

Mockup:

I've often fished values out of large opaque structures with this that would be a pain to do without it

jnvsor:
Fixes #175 

Builds an "Access path" to any accessible variables debugged - eg `$var->prop['index']->method()`

Doesn't show paths to non-accessible members - for standard properties and methods this means recursively checking whether it is accessible.

For public static methods and constants paths are always shown (As they are based on the class not the variable) and when debugging `$this` private and protected members are also shown.

The only issues I can see are:
1. It's currently showing access paths to private members on parent classes from `$this`. This can't be helped inside the decorator, that information isn't passed down in the first place.
2. It takes up some space. It might be prudent to add another button next to "Open in new window" that toggles the access path, but this patch deals with getting it working rather than the UI.


jnvsor:
I've been updating my branch over the past few days and now it's pretty much done. Access path is hidden under a new button on the right, and correctly determines accessibility for everything except private parent class members.

To fix this last (Small and irrelevant) issue would require rewriting the parser to use `var_dump` instead of `var_export` so it's not worth the effort, but looking at `kintParser.class.php` it seems there's an awful lot of markup that should be moved to decorators.

In an ideal world the parser would return data and only data (And lots of it) and all output would be left over to decorators (See https://github.com/jnvsor/kint/commit/0eca0ad65d41ea73c12511ac161eab543eaa4391#diff-191d267d8e9909fb59c16f6c6cff1decR24 for an example where you have to trim quotes from something hardcoded in `kintParser`)

But I digress - it's now pretty much finished:

>

nalipaz:
I am testing this in a drupal 8 installation under the devel module and it is working beautifully.

joachim-n:
Seconding that, this works great with Drupal's Devel module.

jnvsor:

PR #182, opened 2016-01-18
Add mode to delay output until end of script

### LINKED ISSUE #50: create option to output everything only after script execution ends

jnvsor:
Note that this won't work if the user implements a system to send headers like `Connection: close` and `Content-length:` in an attempt to keep the script running after rendering becuase it won't output until the end of the script.

Fixes #50


jnvsor:
Out of scope

PR #187, opened 2016-02-25
Suggested feature to exclude certain classes from being parsed and dumped

tinnerdxp:
Can probably do with some snagging - but does the job at least for me.

Example use with Symfony2:
$_kintSettings['excludeClassInstances'] = ['\Symfony\Component\DependencyInjection\Container'];

and then the DIC doesn't get dumped which speeds things up drastically.

### DIFF
--- a/Kint.class.php
+++ b/Kint.class.php
@@ -48,6 +48,7 @@ class Kint
 	public static $maxLevels;
 	public static $theme;
 	public static $expandedByDefault;
+	public static $excludeClassInstances;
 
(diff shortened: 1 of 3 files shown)


raveren:
Hey, thanks for the contribution!

I'm away from development for some time, but this functionality was already thought of: just add another custom parser to

https://github.com/raveren/kint/tree/master/parsers/objects

that would return minimal info about the huge symfony class(-es). If you create a PR with the symfony object parser, I'll accept it

PR #188, opened 2016-02-26
Add ability to exclude certain class instances from being dumped

tinnerdxp:
As per:
https://github.com/raveren/kint/issues/185

Example use with Symfony2:
$_kintSettings['excludeClassInstances'] = ['\Symfony\Component\DependencyInjection\Container'];

and then the DIC doesn't get dumped which helps a lot.

### DIFF
--- a/Kint.class.php
+++ b/Kint.class.php
@@ -48,6 +48,7 @@ class Kint
 	public static $maxLevels;
 	public static $theme;
 	public static $expandedByDefault;
+    public static $excludeClassInstances;
 
 	public static $cliDetection;
 	public static $cliColors;
--- a/config.default.php
+++ b/config.default.php
@@ -90,5 +90,7 @@ $_kintSettings['cliDetection'] = true;
  */
 $_kintSettings['cliColors'] = true;
 
+/** @var array - allows to exclude certain classes from being parsed, must be a list of fully qualified domain names */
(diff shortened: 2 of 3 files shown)


tinnerdxp:
Hope this is what you meant?
Thanks.

raveren:
I'm sorry, but this is a bad approach for this, I'm leaving the related issue open, but I can't accept this PR.

tinnerdxp:
Ok. Any pointers on what would be the correct way?

raveren:
If there was an easy way, I would have done that already :) What's clear, this should be an inbuilt feature consistent with the rest of the codebase, while your solution (even if it covers the desired functionality well) relies on an unrelated feature and not only does the code become less self explanatory (i.e. surprising), but that whole feature (custom object parser) becomes burdened and more difficult to change, improve or refactor.

I have added a maximum complex item count in my private branch so Kint would not display the humongous framework objects, I'm still testing it out though, but it would fulfill part of the real reason this issue exists...

jnvsor:
I mentioned I was working on a large rewrite. It's delayed by other things. Once it's done this should be doable with a parser plugin (Similar to the custom data/object handlers in kint now, but easier to use from the outside)

tinnerdxp:
Thanks for the explanation guys - makes sense. Looking forward to the release of the feature.

PR #192, opened 2016-04-21
Get mini trace if called through eval

jnvsor:
Currently I'm setting the file and line of the `$callee` to the `call_user_func` call, but simply returning an array of empty values with the mini trace at the end is also a possibility

### DIFF
--- a/Kint.class.php
+++ b/Kint.class.php
@@ -401,7 +401,8 @@ class Kint
 		}
 		$callee = $step;
 
-		if ( !isset( $callee['file'] ) || !is_readable( $callee['file'] ) ) return false;
+		if ( !isset( $callee['file'] ) || !is_readable( $callee['file'] ) )
+			return array( null, null, null, null, $miniTrace );
 
 
 		# open the file and read it up to the position where the function call expression ended


raveren:
Hello there, and thanks for an another contribution!!!

Could you please explain what this PR does in more detail? Also, if possible, give an example of how to achieve the undesired output that your fix makes work, thank you :)

jnvsor:
Actually never mind, my case was a smaller case of the real issue: The minitrace doesn't show when called through `eval`:

`eval("d('debug', 'this');");`

Considering this larger case the second method is the way to go, and I'll be pushing a new patch that works better (And is cleaner too! Yay!)

As to why you would ever want to use eval - some templating systems _cough_ _twig_ _cough_ use it to run their templates.

raveren:
Eval is not necessary evil in such tools, if the situation forced me to, I would use eval in Kint, and I can imagine situations where Twig developers where forced by PHP to do it as well.

I'm closing this issue, right?

jnvsor:
The problem is that the mini trace doesn't show if kint is called through eval:

``` php
function test(){
    d("test");
    eval('d("test");');
}
test();
```

PR #196, opened 2016-04-24
Show caller info on collapsed mini trace in eval

jnvsor:
See #192


raveren:
I merged without checking again, this is a bugfix for the other one?

jnvsor:
More like a feature (It worked before but was empty next to the plus when collapsed) but I just thought of it after you merged

raveren:
ok, phew

PR #197, opened 2016-04-25
Bugfix: Check aliases case sensitive

jnvsor:
The aliases property can help kint wrappers keep clean output, but currently it's case insensitive. (Did you find a quirk in a specific version of PHP's debug_backtrace?)

This means it will incorrectly identify `test` `Test` `TEST` and `tEsT` (Among others) as being an internal function call, breaking modifiers, the minitrace, and full traces.

### DIFF
--- a/Kint.class.php
+++ b/Kint.class.php
@@ -603,14 +603,18 @@ class Kint
 	{
 		if ( isset( $step['class'] ) ) {
 			foreach ( self::$aliases['methods'] as $alias ) {
-				if ( $alias[0] === strtolower( $step['class'] ) && $alias[1] === strtolower( $step['function'] ) ) {
+				if ( strtolower( $alias[0] ) === strtolower( $step['class'] ) && strtolower( $alias[1] ) === strtolower( $step['function'] ) ) {
 					return true;
 				}
 			}
-			return false;
 		} else {
-			return in_array( strtolower( $step['function'] ), self::$aliases['functions'], true );
+			foreach ( self::$aliases['functions'] as $alias ) {
+				if ( strtolower( $alias ) === strtolower( $step['function'] ) ) {
+					return true;
+				}
+			}
 		}
+                return false;
 	}
 
 	private static function _parseTrace( array $data )


willzyx-dev:
:+1:

willzyx-dev:
Another viable approach could be to use strtolower() both for alias and for step. In this way the problem is solved and you maintain backward compatibility with application that use kint aliases

jnvsor:
That doesn't solve the problem if you actually _have_ classes or functions that differ by case alone. I'm working on a large revamp of kint internals, enough to justify a major version bump. That should open the floodgates for all the improvements and fixes I've stockpiled.

willzyx-dev:
> That doesn't solve the problem if you actually have classes or functions that differ by case alone. 

Not sure to understand correctly this affermation.. functions classes and methods are (sadly) case-insensitive in php

jnvsor:
Oh wow. Holy shit. I never actually considered that. _That's horrible!_

Then you're correct that double `strtolower` would do the trick, but I'm busy with other patches and raveren's nowhere to be found so I'll leave it for a bit.

raveren:
Hey, just an update, sorry for being AWOL. I no longer work with PHP as my main language and I've been under quite a bit of stress lately and am now only shaking it off, I will continue to contribute to Kint, but I need a week or two to catch up with all of my worries. Kint is in the top priorities, don't worry and thanks for the hard work meanwhile.

As for this PR, the solution is not complete, you need to deep-lowercase the whole `Kint::$aliases` array and _then_ compare with the data from trace.

jnvsor:
How's that?

raveren:
Oh, sorry my bad, I must have looked too hastily, if you could get rid of the L63-L64 unnecessary change, that would be awesome.

jnvsor:
Done.

It might be more readable to replace the loop with `array_map( 'strtolower', self::$aliases['funtions'] )` - should I push that instead?

raveren:
I am now looking at this change again and seeing it does not change anything but fix a possible slight user misconfiguration. I agree that it makes Kint a tiny bit more user friendly, but I don't understand any of the arguments:

> This means it will incorrectly identify test Test TEST and tEsT (Among others) as being an internal function call, breaking modifiers, the minitrace, and full traces.

Can you provide a use case to reproduce this, I'm missing something, sorry

jnvsor:

Issue #193, opened 2016-04-22
'Called from' line shows Kint's code, rather than the caller


joachim-n:
When I output a variable with Kint, the 'Called from' line says:

> Called from <ROOT>/mysite/web/modules/devel/kint/src/Plugin/Devel/Dumper/Kint.php:51 [Drupal\kint\Plugin\Devel\Dumper\Kint->export()]

It should show me the file where I actually requested the output.

The full backtrace starts like this:

```
<ROOT>/mysite/web/modules/devel/src/DevelDumperManager.php:85 [Drupal\kint\Plugin\Devel\Dumper\Kint->export()]
<ROOT>/mysite/web/modules/devel/src/DevelDumperManager.php:95 [Drupal\devel\DevelDumperManager->export()]
<ROOT>/mysite/web/modules/devel/devel.module:345 [Drupal\devel\DevelDumperManager->message()]
<ROOT>/mysite/web/modules/devel/devel.module:375 [dpm()]
<ROOT>/mysite/web/core/modules/views/src/Plugin/views/relationship/RelationshipPluginBase.php:180 [dsm()]
```

This is using Drupal's Devel module. As you can see, there are a few function calls between the place where I call it using dsm() and the actual call to Kint.

Is there something that Devel should be passed in to Kint->export() to tell it how many items to remove from the backtrace()?

(Corresponding issue on Devel: https://www.drupal.org/node/2643392#comment-11060295)

raveren:
Unfortunately I am not the owner of the drupal fork and I don't even know what the `export` method does, but `Kint::_stepIsInternal` method is responsible for defining how many steps to roll back and show the true caller in the trace...

joachim-n:
Oh heck, is the Kint code inside that module a fork? Urgh :(

raveren:
That's actually my fault, I was unresponsive to their change requests, but it required a heavy rewrite so I didn't find the time sadly...

jnvsor:
> That's actually my fault

Not really. I've been working on a kint extension to bolt/bolt and all I have to do to get decent traces is add my entry function to `\Kint::$aliases`

@joachim-n I suggest you raise this as a bug report on the devel module

joachim-n:
So is the right way to do this just to add to Kint::$aliases before calling Kint::dump(), like this:

```
      Kint::$aliases += [
            'dsm',
            'dpm',
            'kprint_r',
      ];
[...]

PR #202, opened 2016-05-21
Make kint minitrace more readable

willzyx-dev:
use light-grey color (#ddd) makes the minitrace very difficult to read in almost all kint themes. 
The only theme in which use #ddd has sense is "solarized-dark"

kint-minitrace


jnvsor:
I can pretty much tell you right now that this isn't going to happen.
1. The text is deliberately grey. if you remove that rule it goes back to being black. It's grey because...
2. ...the minitrace is supposed to be tertiary information. If you want a backtrace there's both a special syntax `d(1)` and method `Kint::trace()` which will both give you far more information.
3. If you want custom colors, you should make your own theme instead of trying to change the default one. (You don't even need the theme, you can throw a more specific selector into the html completely outside of Kint)

willzyx-dev:
@jnvsor Thanks for the speedy reply
1. I can agree with you that as tertiary information minitrace need a different color.. but use #ddd make the minitrace unreadable.
2. I know that kint provides ::dump(1) and other shorthands for getting the backtrace but is not what I'm looking for. I would like to be able to read the minitrace that kint provides
3. The point here is that use light-grey color (#ddd) makes the minitrace very difficult to read and this is an accessibility issue.

jnvsor:
> very difficult to read and this is an accessibility issue

Unless you're using the default theme on a black background. At which point making it darker would be an accessibility issue.

People don't always use the same theme. This is why themes exist in kint - you should use them. (Or just chuck a single css rule into your output: `.kint footer ol>li{color:red}`)

In actual use I almost never use the minitrace. Does this have something to do with devel's not correctly setting a Kint alias so the minitrace shows the wrong stack frame from #193 ?

willzyx-dev:
> Unless you're using the default theme on a black background. At which point making it darker would be an accessibility issue.

of course :) but I think that sites that use black background are a minority compared with sites that use a white/light color background

> In actual use I almost never use the minitrace. Does this have something to do with devel's not correctly setting a Kint alias so the minitrace shows the wrong stack frame from #193 ?

No this PR is absolutely not related to #193. For that problem we have updated kint to the latest tagged version and we are working for use \Kint::$aliases + \Kint::$returnOutput as you suggested

jnvsor:
That's not going to work - the latest tagged version was 9 months ago and you already had that version. Haven't had a new tag released in ages.

But back on topic - perhaps you should rework this PR to give the minitrace a background. That way you can guarantee that it actually does improve readability

willzyx-dev:
yes, adding a background color to the minitrace container can be a viable solution. But how to choose a decent default background? use white by default doesn't seems the right approach.. probably you need a default per theme background. have you any suggestions?

minitrace-container

something like this?

jnvsor:
I think default white is fine - perhaps at 90% background opacity so the underlying theme shows through a bit. Throw in rounded bottom corners and inline-block to make it not full width and you should be done:



**Edit:** Or that, that should work too

willzyx-dev:
oh I like the default white-90%-opacity background!! probably make the minitrace container full width can give a little more symmetry the the output :)

jnvsor:
It's full-width by default. I was thinking about large debug outputs when I did that, there's an awful lot of empty space to the right of the mini trace compared to the rest of the output, and it just feels less "wrong" in my gut when it's not full width

willzyx-dev:
> It's full-width by default. I was thinking about large debug outputs when I did that, there's an awful lot of empty space to the right of the mini trace compared to the rest of the output, and it just feels less "wrong" in my gut when it's not full width

It make sense :+1:  used inline-block for minitrace container

jnvsor:
You should use `background: rgba(...` instead of `opacity` as opacity makes the contents fade out too

Also, I'm not sure solarized-dark even needs specific changes now - how does the white minitrace look?

willzyx-dev:
The css changes was done in base.less and affect all themes. In solarized-dark.less (solarized-dark is the only theme that change the body background) the css for the container is overridden and setted to none

jnvsor:
Ah, didn't realise it changed the background. Still, the opacity and border-radius lines are moot - and when you've done that try squashing them all into one clean commit - easier to review

willzyx-dev:
> Still, the opacity and border-radius lines are moot

why do you think that these lines are moot? Do we need to add fallback for browsers that don't support these properties?

jnvsor:
You should squash your commits. You've got duplicate CSS in between them, it's confusing at best.

jnvsor:
Perfect. :+1:

aaronbauman:
+1 for this, thanks!

raveren:
Hello, thanks for the hard work, the mini-trace is definitely underdeveloped, but was so useful in everyday work, I chose to include it as it was. It currently still has issues with keyboard navigation ("what?" yes, press `d` and you can expand/collapse nodes, but navigating the minitrace throws js errors :/)

Anyway, why the rounded border and the inline display again? The default theme is an axiom, it was designed by a professional, I dare not change it myself. Let's keep this PR for the trace color only and I'll merge it instantly.

jnvsor:
That was when it was a transparent background. Full width opaque background with nothing in it doesn't look too good (See for yourself)

willzyx-dev:
Reverted the minitrace container background and the inline display

jnvsor:
@willzyx-dev He didn't say you should get rid of the background - that's a legibility loss IMO

raveren:
I tried the changes & the themes look very readable now, @jnvsor is there a real issue I'm not aware now? 

Note, I do know aante-light wasn't compressed by mistake, I'll push a quite big batch of changes later this week that fixes many issues and this is one of them...

jnvsor:
@raveren In the old version, if you used the original theme on a black background you could read the trace. Now the background is gone the text will be completely illegible on a black background (I think @willzyx-dev did a blind revert and forgot about the background)

raveren:
hey, @willzyx-dev, how's it going, I hope we didn't discourage you with our nitpicking :) Thanks for the initiative, what's your take on the background thing? :)

PR #206, opened 2016-07-13
Fix jQuery on https access due mixed content blockade

ottnorml:


jnvsor:
Shouldn't this be a protocol relative url? (IE `//code.jquery.com/jquery-1.9.1.min.js`)

ottnorml:
@jnvsor You're right. This would be a better solution.

ottnorml:
Changed it and created a [new pull request](//github.com/raveren/kint/pull/207).

PR #209, opened 2016-08-05
Enhanced fonts to readable sizes, better padding and content wraps for responsive layouts.

prezire:
Here's a screenshot http://icecream.me/2756188727b7abb931ade7fbb83a36bf.


raveren:
Hello and thanks for the effort, however I am not sure of some of the changes. For example you are reverting the footer color just recently changed in #202, you're also committing `compiled/kint.js`, which you didn't alter, just recompiled. I'm under the impression you based your changes on an older commit of Kint than HEAD and you have some unrelated changes in this PR (which are basically reverts).

For the rest of the changes, they seem subjective, and I'd like to see a problem each of them solves (refer to the excellent description of #202), for example I don't remember the reasoning for `white-space : nowrap;` which you change to `normal`, but I am sure this was made after hours of experimenting..

prezire:
Hi raveren. Thanks for the tips. I based out from the latest branch. I consistently use the same git flow from the company I work with and when I contribute to CodeIgniter. Apologies if I missed something in the commit.

The reason for changing `white-space` is for the debugger to work nicely with mobile devices (especially to prevent horizontal scrollbars). Please feel free not to accept any changes if there are a lot of issues. I hope though, the fonts would get larger based on the screenshot I provided. I'm nearsighted but I can't even see the stack trace properly :)

raveren:
ok, that is a valid complaint and I can I agree the font size might be too small and I would accept a PR that changes that - and until I can see reasoning for other changes - only that. Meanwhile I'm closing this, as the PR has way too many unrelated changes...

Issue #219, opened 2016-11-03
$tabularData on examples/overview.php


DRSDavidSoft:
**Kint Version:** (Latest as of writing this )

**How it is supposed to look:**
![image](https://cloud.githubusercontent.com/assets/4673812/19964362/8d6fe7ce-a1d6-11e6-8c83-255ed753a348.png)

**How it does look right now:**
![image](https://cloud.githubusercontent.com/assets/4673812/19964320/550be9be-a1d6-11e6-9f1c-c91efcbe1483.png)

**Proposed fix:** 
Line 118: `for ( $i = 1; $i < 9; $i++ ) {`
Line 127: `if ( $i % 3 == 0 ) {`

jnvsor:
If you've already fixed it, leave a pull request

DRSDavidSoft:
@jnvsor, Created a PR on #220.

jnvsor:
Tabluar data has been moved into a plugin in Kint 2.0, which has a stricter take on tables to begin with. (Allows empty, but not missing keys) Is this issue still relevant?

DRSDavidSoft:
I should take a look at your unit tests first before giving any definite conclusions.

And BTW, I'm really excited about Kint 2.0, it promises way more flexibility and seems to be more powerful than Kint 1.x, however I think that while the new plugin is stricter, it should still allow missing keys, at least using a pre-initialization config.

Do you agree on this?

jnvsor:
Perhaps. But the logic for missing keys gets a lot harder to reason with: Hence this issue

> I should take a look at your unit tests first before giving any definite conclusions.

Hah! Good one!... Oh... You're serious?

Joking aside, Kint wasn't decoupled enough to unit test before 2.0 and I've spent all my time cramming functionality in there. Unit tests are on the todo list (And I'm working on TravisCI setup [in my personal repo](https://github.com/jnvsor/kint/commit/1486ad37e3ddb47f87acfd20938df06968bb1453)) but I expect 2.0 to ship before coverage hits double digits.

DRSDavidSoft:
Let's see how the 2.0 revision turns out before dealing with this. I can only ask that you keep an option for missing keys in your plans, so we could code a way around the hardness.

It's nice that there will be unit tests, it's a necessity in my opinion ;)
May I suggest combining them with the examples? That's what a lot of repos do, for obvious reasons.

[...]

PR #220, opened 2016-11-03
Fixed $tabularData on overview.php

DRSDavidSoft:
Apparently, In order for the latest versions of Kint to show 2-d arrays in tabular-view, no columns should have more than 50% empty cells .

### DIFF
--- a/examples/overview.php
+++ b/examples/overview.php
@@ -116,10 +116,11 @@ $user->setCreatedDate( new DateTime( '2013-10-10' ) );
 $userManager = new UserManager();
 
 for ( $i = 1; $i < 12; $i++ ) {
+	$number = str_pad($input, 2, "0", STR_PAD_LEFT);
 	$tabularData[] = array(
-		'date'        => "2013-01-0{$i}",
+		'date'        => "2013-01-{$number}",
 		'allowed'     => $i % 3 == 0,
-		'action'      => "action {$i}",
+		'action'      => "action {$number}",
 		'clicks'      => rand( 100, 50000 ),
 		'impressions' => rand( 10000, 500000 ),
 	);
--- a/examples/overview.php
+++ b/examples/overview.php
@@ -116,7 +116,7 @@ $user->setCreatedDate( new DateTime( '2013-10-10' ) );
 $userManager = new UserManager();
 
 for ( $i = 1; $i < 12; $i++ ) {
-	$number = str_pad($input, 2, "0", STR_PAD_LEFT);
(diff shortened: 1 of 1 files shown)


DRSDavidSoft:
Fixes #219.

DRSDavidSoft:
Is this PR still relevant for Kint 2.0 examples? Or the new plugin replaces this?

jnvsor:
The new plugin replaces it. It's also more strict than the old behaviour so it'll likely make your issue in #219 worse, though you can mitigate that by writing your own plugin for it.

If you don't find any massive explosive issues with it, mind if I close the PR and issue?

DRSDavidSoft:
I don't mind at all, but IMO it's more appropriate if you push it to the 1.0 branch for historical versions (some future visitors might still be wanting to use the older version of kint.)

It's just an example script, after all.

jnvsor:
Ah true, I thought this was tweaking behaviour not an example. Though I do spy a bug right off: The day in the dates will go up to `2013-01-011` now. Lower the condition to 10 or so and it should be fine

DRSDavidSoft:
You have eagle eyes! I hope you accept this solution, I tried to keep it still readable.

examples/overview.php
```diff
@@ -115,16 +115,17 @@ public function ensureUser() { $this->user->ensure(); }
 $user->setCreatedDate( new DateTime( '2013-10-10' ) );
 $userManager = new UserManager();
 
-for ( $i = 1; $i < 6; $i++ ) {
+for ( $i = 1; $i < 12; $i++ ) {
+	$number = str_pad($input, 2, "0", STR_PAD_LEFT);
```
jnvsor: I spy with my little eye, copy pastey pastey :)

jnvsor (changes requested):

examples/overview.php
jnvsor: In case you still don't see it: `$i != $input`

DRSDavidSoft: That's actually a misplaced auto completion. That's what happens if I hurry to commit :D

DRSDavidSoft (commented):
Done. Sorry about multiple commits!

DRSDavidSoft:
Everything good for the PR now?

jnvsor:
No problem - I'll squash them in the end. Just let me test this quickly

jnvsor (approved):

jnvsor:
Squashed as 425c6127b54e47934f36d69519342c94d2d1c911

PR #223, opened 2016-12-20
Bug fixed: error on default timezone not set

lazharous:


lazharous:
date_default_timezone_set('UTC') should be moved few lines above as the DateTime() is called previously.

PR #225, opened 2016-12-21
Vim navigation keys (& keyboard error fix)

neitanod:
Use "hjkl" as arrow keys (for Vim fanatics)
Use "j" as an alternative to enable keyboard navigation (makes sense for Vimmers too).
Fixed error when trying to expand "Called from..." using keyboard navigation


raveren:
Hello and thank you for the support - and the work you put into your two PRs!!! 

The bugfix is but a workaround unfortunately, the perfect case scenario would be to make the shortcut actually expand the footer, however I haven't gotten around to that - still it's better than previously, so it's acceptable as a change. I'll add the few other nitpicks in the code review :)

view/base.js
```diff
@@ -378,9 +381,9 @@ if ( typeof kintInitialized === 'undefined' ) {
 			kint.toggle(kintNode);
 			kint.fetchVisiblePluses();
 			return false;
-		} else if ( keyCode === 39 || keyCode === 37 ) { // ARROW LEFT/RIGHT : respectively hides/shows and traverses
+		} else if ( keyCode === 39 || keyCode === 37 || keyCode === 76 || keyCode === 72 ) { // ARROW LEFT/RIGHT : respectively hides/shows and traverses
```
raveren: this line is too long, I think 120 is a sane maximum

view/base.js
```diff
@@ -338,8 +341,8 @@ if ( typeof kintInitialized === 'undefined' ) {
 			, shiftKey = e.shiftKey
 			, i = kint.currentPlus;
 
-
-		if ( keyCode === 68 ) { // 'd' : toggles navigation on/off
+		if ( keyCode === 68 || (keyCode === 74 && j_toggles) ) { // 'd' or 'j' : toggles navigation on/off
```
raveren: So 'j' toggles navigation as well as navigate - but how do you get out of kint navigation mode and just want to scroll the page with arrow keys? 'd'? That's a bit counter-intuitive, don't you think? I'm not a VIM buff, so what's your oppinion?

view/base.js
```diff
@@ -352,9 +355,9 @@ if ( typeof kintInitialized === 'undefined' ) {
 
 			if ( keyCode === 9 ) { // TAB : moves up/down depending on shift key
 				return kint.keyCallBacks.moveCursor(shiftKey, i);
-			} else if ( keyCode === 38 ) { // ARROW UP : moves up
+			} else if ( keyCode === 38 || keyCode === 75 ) { // ARROW UP : moves up
```
raveren: This as well as other places should reflect your change in the comments, cause keyCode 75 is highly unreadable. Please observe the line limit as well.

view/base.js
```diff
@@ -434,4 +437,4 @@ function clg( i ) {
 	if ( !window.console )return;
 	var l = arguments.length, o = 0;
 	while ( o < l )console.log(arguments[o++])
-}
\ No newline at end of file
+}
```
raveren, line 447: This is a nitpick, but changes to spacing mess up git blame, it's too easy to avoid it here to not to :)

view/base.js
```diff
@@ -329,6 +331,7 @@ if ( typeof kintInitialized === 'undefined' ) {
 	}, false);
 
 	// keyboard navigation
+	var j_toggles = 1;
```
raveren: Nitpick: j_toggles is an undescriptive variable name (I know, the file is not that readable already), but it a) indicates it's a counter just by it not having a boolean value b) `navStartedByJ` or something similar would adhere to camelCase convention of the whole file and describe the purpose of the variable better

raveren (changes requested):
sorry for being a bit anal :)

neitanod:
The fix works fine in my machine.  Give it a try  :)
It stops the "do ... while" from diverting to other element when current one is "footer", so it expands both on click and on key press (I tried <right_arrow>, <enter> and <L>).

view/base.js
neitanod: This change to spacing was accidental (initially) and I was trying to get it back to how you got it, but failed, apparently.  :P

view/base.js
neitanod: I did split it originally, and that's when I realized that you use tabs for indentation.  I didn't really know how to indent such a half line correctly.
I'll give it a try, and you please correct me if I get it wrong.

view/base.js
neitanod: The ESC key would be the "Vim way" to get out of navigation mode, I think.

neitanod: The boolean just says if keyboard navigation is active already or not, so I'm about to push it with the name _keyboardNavigationIsActive_

jnvsor:
> make the shortcut actually expand the footer, however I haven't gotten around to that

That's working in 2.x. I'll be pushing my latest commits to jnvsor and Kint-PHP soon

@neitanod I made a quick 'n dirty hjkl implementation in 2.x - is it really a good idea to turn on keyboard nav with j though?

jnvsor:
I added the hjkl navigation to kint in 2.x-dev, and fixed the error before then. I'm not convinved j to activate in addition to d is worth the effort. Could you try it and see how it feels?

Issue #222, opened 2016-11-25
PHP7 - Warning: A non-numeric value encountered


turbopixel:
We use PHP 7.1 for a few days. Every Kint Dump print a warning: `Warning: A non-numeric value encountered in ...../class/SearchCore.php on line 336`

Line 336: `-sd( $data );`

$data is an array().

**Details**
PHP Version: PHP 7.1.0RC6
php-fpm + nginx
php_ini: error_reporting = 32767
Linux version 4.2.0-36-generic (gcc version 4.8.2 (Ubuntu 4.8.2-19ubuntu1) ) #41~14.04.1-Ubuntu

jnvsor:
There is no `SearchCore.php` in kint. Your problem lies elsewhere

turbopixel:
I put `-sd( $data )` in my own file calling *SearchCore.php*. And this returns the php error `Warning: A non-numeric value encountered` ..

Code:

```
class SearchCore{
	
	// ...

	public function loadQuery($data){

		-sd( $data );

	}

	// ...

}

jnvsor:
Ah that makes more sense. I don't have a 7.1 RC here to test but if I had to guess I'd say 7.1 is cracking down on invalid type casts and Kint's modifiers are running foul of that.

Can you confirm if this happens with `sd($data)`? (No `-`)

turbopixel:
(I'll test this on next monday)

turbopixel:
**Old:**
The problem only occurs in our company framework.

**Edit 03:04PM:**
Yes `-d($array)` create the warning. `d($array)` not.

PR #226, opened 2016-12-21
Return provided data

neitanod:
Hello @reveren  :)

I am Sebastian, author (in 2010) of a very similar tool (but very modest in comparison):

[http://inspect.ip1.cc](http://inspect.ip1.cc)  (It's in spanish but you can see it's basically the same)

I would really love to ditch my old "inspect" class and start using your fantastic Kint!

But there's a feature in my tool that Kint is missing, and I think it would be a nice improvement.

By default Kint returns an empty string.  To me, returning the same value received as first parameter makes a lot more sense.  That would allow us to inspect mid-expression values:

if ( !is_readable( $cssFile = $baseDir . **d(** _Kint::$theme_ **)** . '.css' ) ) {

Modifiers will still replace the returned value as intended.

What do you think?

### DIFF
--- a/Kint.class.php
+++ b/Kint.class.php
@@ -220,7 +220,7 @@ class Kint
 		if ( $trace ) {
 			$output .= call_user_func( array( $decorator, 'decorateTrace' ), $trace );
 		} else {
-      $dataArray = func_num_args() === 0
+			$dataArray = func_num_args() === 0
 				? array( "[[no arguments passed]]" )
 				: func_get_args();
 
--- a/Kint.class.php
+++ b/Kint.class.php
@@ -220,11 +220,11 @@ class Kint
 		if ( $trace ) {
 			$output .= call_user_func( array( $decorator, 'decorateTrace' ), $trace );
 		} else {
-			$data = func_num_args() === 0
+      $dataArray = func_num_args() === 0
 				? array( "[[no arguments passed]]" )
 				: func_get_args();
 
-			foreach ( $data as $k => $argument ) {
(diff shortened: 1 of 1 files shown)


raveren:
That is a nice idea! I have to think about it for a while whether it might have unexpected side results, but I'll most probably merge it. What do you reckon, @jnvsor?

jnvsor:
It's a nice idea but I'd suggest sticking it behind a feature flag.

More importantly, if #222 wasn't just a false alarm we may have to get rid of modifiers entirely so I'd wait until php 7.1 release before pulling this.

Lastly there's the question of whether this may cause unexpected issues if implemented generically (Specifically if you pass a huge array to it it's going to get copied not referenced so the helper functions would need extra code)

I'll stick it on my wanted features list

**Edit:** Nope this won't work. Try using modifiers with an array as the parameter:

> **Fatal error**: Unsupported operand types in...

So in the end this will still have to wait until we see 7.1 behaviour. If we have to get rid of modifiers anyway this would be a nice feature

jnvsor:
In PHP 7.1 modifiers will make a nice fireball if you're not returning a compatible data type. Implementing this would require us to throw away the modifier system.

I'm not massively against that (It feels a bit hackish in retrospect) but I'm not going to remove such a useful feature. I suggest you instead implement your own wrapper. Example:

```php
function r(...$args)
{
    Kint::dump(...$args);
    return count($args) == 1 ? reset($args) : $args;
}
Kint::$aliases[] = 'r';
```

PR #235, opened 2017-04-24
Update _base.scss

marijaninjo:
trailing > removed


jnvsor:
That's not an accident:

```scss
a > {
  b { color: red; }
}
a {
 b { color: white; }
}
```

Becomes:

```scss
a > b { color: red; }
a b { color: white; }
```

So basically: The trailing `>` is significant

Issue #256, opened 2017-07-31
how to add new function to helpers?


Mehrdad-Dadkhah:
hi

tanx for good package

how can I add new function as helper function without edit source file of library (init_helpers.php)?

jnvsor:
Have a look at the [documentation](https://kint-php.github.io/kint/advanced/#helperfuncs)

Mehrdad-Dadkhah:
@jnvsor 
yes I do that before and I test it  in laravel.
my problem is that in composer autoload_files.php init_helpers.php load before laravel helpers and so it's work fine, but when I define dd() function and add it to Kint::aliases I have redeclare function error because laravel helper define dd() function before and my target is to overwrite it.
If I can overwrite init_helpers.php and in init.php can load another init_helpers.php (with some configs) my problem resolve ...

jnvsor:
No, that won't resolve your problem, it'll just cause a fatal error when laravel tries to define dd instead. Your conflict is an issue with laravel, not Kint. I don't know if there's a way to remove dd from laravel without creating a merge hell (I doubt it) but in either case your best option is to use a different name. (Kint 1 added `ddd` specifically because of the conflict with laravel)

**Edit:** A quick grep shows modern laravel defines dd [here](https://github.com/laravel/framework/blob/5.4/src/Illuminate/Support/helpers.php#L517) - It's got a defined check so if you really want your dd defined first you'll want to stick it in your composer [autoload](https://github.com/laravel/framework/blob/master/composer.json#L89) or some other part of the code that's called earlier than this code.

Mehrdad-Dadkhah:
Thanks you for reply @jnvsor 

Yes, I know that my conflict is with laravel and because in composer auto load helpers.php of laravel load faster than my helper.php (where I define dd function) and my target is overwrite laravel dd. But if I can change path of init_helpers.php of kint and load another init_helpers.php my problem will be fixed.
for example if I can change KINT_DIR as result it will load another init_helpers.php (my helper file instead of kint helper file)

I point to here:

```
require_once KINT_DIR.'/init_helpers.php';
````

If we make KINT_DIR optional, Kint will be more flexible.

jnvsor:
No. `KINT_DIR` is integral to... Well a whole lot of stuff. I won't bore you with the details but that's not the way to do this. Even if it was, how exactly were you planning on doing that? Set a global or define before calling composer? Because as soon as composer is required the autoload on laravel is going to trigger too, and without composer you don't have Kint either.

[...]

PR #257, opened 2017-08-01
define KINT_DIR if not defined before, change to can load another init_helpers.php

Mehrdad-Dadkhah:
hi
simple and little changes related to [issue 256](https://github.com/kint-php/kint/issues/256) 

now we can define KINT_DIR before require composer autoload and make it compatible with us use like overwrite laravel dd to show use power of Kint.

### DIFF
--- a/init.php
+++ b/init.php
@@ -4,10 +4,6 @@
  *
  * https://github.com/kint-php/kint
  */
-if (defined('KINT_DIR')) {
-    return;
-}
-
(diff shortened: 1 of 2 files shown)


jnvsor:
Like I said in #256 `KINT_DIR` is important for lots of stuff and there's no way to solve your very specific issue (Which incidentally has *nothing* to do with Kint) that isn't a dirty hack.

The only surprising thing about this commit is that one of the tests *didn't* fail. I've since updated the test.

PR #282, opened 2018-07-03
Update README.md

Niellles:
Added --dev option to composer command, since this package should only be required in a development environment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ require 'kint.phar';
 ### Or, if you use Composer:
 
 ```bash
-composer require kint-php/kint
+composer require kint-php/kint --dev
 ```
 
 ## Usage


jnvsor:
I'm not sure that's true - you can enable it conditionally to debug in live. Yes that's a bad idea but there's too much legacy code out there not to.

But agreed - it should be `--dev`

Let me figure out why CI is complaining and merge it afterwards

Niellles:
If you're enabling it in a live environment you may as well install dev packages for the time being and then later remove them with `composer update --no-dev`. I would argue that that's better practice than just disabling it like the readme says:

> The best practice is to enable Kint in a development environment only - so even if you accidentally leave a dump in production, no one will know.

The legacy code that requires this package in a live environment probably won't break because of this; it's already installed. Strictly speaking this doesn't really break anything unless people choose to run the command without understanding what the `--dev` option does.

jnvsor:
> I would argue that that's better practice than just disabling it like the readme says:

The problem is that if you accidentally leave a dump in production and an update removes the package (Which is probably installed on your dev server) now you've gone from 'no-op debug statement' to 'fatal error calling nonexistant function'

> Strictly speaking this doesn't really break anything

Yes that's why I said I'd merge it - but I'm afraid you'll have to wait a bit: Someone decided to play piano with the breakers in my flat and I'm stuck rebuilding my machine before I can even think about this

jnvsor:
I've fixed CI, can you rebase so I can merge when CI goes green?

Niellles:
I did the patch via GitHubs web-interface and have rebased via the same interface, CI doesn't seem to catch it though.

jnvsor:
Merged

PR #299, opened 2019-01-25
Cleaned up warnings and provided some error checking in build.php

deanhouseholder:
Added a try/catch block to minimize PHP fatal error output while retaining the message.

### DIFF
--- a/build.php
+++ b/build.php
@@ -28,12 +28,22 @@ use Symfony\Component\Finder\Finder;
 
 require_once __DIR__.'/vendor/autoload.php';
 
-\mkdir(__DIR__.'/build');
+if (!\file_exists(__DIR__.'/build')) {
+    \mkdir(__DIR__.'/build');
+}
 
 $outpath = __DIR__.'/build/kint.phar';
+if (\file_exists($outpath)) {
+    \unlink($outpath);
+}
 
-\unlink($outpath);
-$phar = new Phar($outpath);
+try {
+    $phar = new Phar($outpath);
+} catch (Exception $e) {
+    echo 'PHP Fatal error: '.$e->getMessage();
+    die;
+}
+
 $phar->setStub('<?php
 /*
  * '.\str_replace("\n", "\n * ", \trim(\file_get_contents(__DIR__.'/LICENSE'))).'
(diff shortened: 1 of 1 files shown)


build.php
```diff
@@ -28,12 +28,22 @@
...
+if (\file_exists($outpath)) {
+    \unlink($outpath);
+}
+
+try {
+    $phar = new Phar($outpath);
+} catch (Exception $e) {
+    echo 'PHP Fatal error: '.$e->getMessage();
+    die;
+}
```
jnvsor, line 45: This try catch is redundant, since the exception will result in a fatal error anyway (And a real one, with the proper exit code, print to stderr, etc)

build.php
```diff
@@ -55,3 +65,5 @@
 $phar = new Timestamps($outpath);
 $phar->updateTimestamps();
 $phar->save($outpath, Phar::SHA512);
+
+echo 'Your kint.phar has been built successfully and is located in the "build" directory.';
```
jnvsor, line 69: Since there's no output there's a future possibility for us to pipe the script to a file instead of writing directly, I think we should just leave stdout empty for now

jnvsor (changes requested):
> I'm going to cancel and submit a new one.

FYI: You can force push to your branch to automatically update the PR, no need to make a new one

raveren:
Is this maybe why our whole build is broken for weeks :DD @jnvsor I don't think we should keep this open, @deanhouseholder thanks for the effort, and sorry we couldn't make use of the code, but people like you is who inspire us to build things like this.

jnvsor:
@raveren No, the build isn't broken. @deanhouseholder I'd like to reopen this, but if you don't feel like finishing it up I'll leave it closed.

deanhouseholder:
I can't understand why you prefer ugly errors to clean ones. Haha

jnvsor:
This gets a bit unixy as far as technicals go, but programs with text input/output generally have 3 streams: stdin (input) stdout (output) and stderr (error)

Why do you have different streams for output and error? Because it's common to connect programs up to eachother. For instance you could pipe a php script's output to another one like so:

```bash
php script.php | php script2.php
```

If you do it like this, only stdout (The output) is piped to script2, and stderr ends up in the terminal. Apache modphp typically sends both stdout and stderr to the browser, so it doesn't matter there.

The second reason is that programs return a number when they end indicating success or failure. A 0 is success and anything else indicates a variety of failures.

```bash
php script.php && php script2.php
```

In this example script2 will only run if script1 succeeded, which it knows by the exit code. A proper fatal error returns a non-0 exit code, which is also used by CI to tell if things succeeded.

In other words by manually exiting the script *exceptions in the build won't be caught by CI* which is a very bad thing.

Long story short: Smarter people than us thought up these systems and they work very well. No touchy :)

PR #302, opened 2019-02-15
Added quick instructions to display Kint inline again.

raveren:


jnvsor:
I like to use "set `...`" in the readme and leave most features for the `gh-pages` branch, but yeah I got a lot of issues when I pushed 3.0 and people were wondering where their dump went :)

raveren:
it just saves you a click each and every time you do a change

Issue #70, opened 2013-06-06
show __toString info on objects header row


winkbrace:
When an object has a __toString() function, get it and display it in the header row. Handy for understanding at a glance collapsed objects.

aljinovic:
Calling methods from an object can cause problems that can not be predicted, since __toString(), like any other method, can execute any code, causing errors or performance issues. If you plan on adding this functionality, please make it optional and off by default in configuration.

raveren:
Sorry, will not be implementing this. I even opted out unserializing applicable strings so as not to trigger `__wakeup` methods. 

Too risky business, wouldn't be comfortable having a ticking timebomb in Kint code.

winkbrace:
Okay, I hadn't thought of it like that. I'll just make a plugin for my own
use then. :)

;} Bas de Ruiter

2013/6/6 Rokas Šleinius notifications@github.com

raveren:
If you go around doing it, it's very simple - just create a new parser in `kint\parsers` directory. You might create a pull request, I'll accept it once I create a parser management system of some kind.

PR #304, opened 2019-03-06
Add smarty plugin

raveren:
Hey, I ran into a foreseen bug #70 while debugging an old system.

[Smarty](https://www.smarty.net/) would kill Kint not only by it's share size of objects, but also since calling its `__toString` would cause exceptions. So I decided to have some fun and wrote a plugin to hit two birds! :)

The process was mostly painless, the way you handled the plugin architecture is not just workable, I'd say it's extremely elegant for what it is!

I did have some counter-intuitive problems to solve - precisely how to display a sub-element of a variable instead of the variable itself; this ended up working
```php
$o = $this->parser->parse($var->value, $o);
```
maybe I should add this as a tip to [the docs](https://kint-php.github.io/kint/writing-plugins/)? 

Also, as I mentioned in the PR code, I stumbled on a bug which I'm too afraid to debug:

Trace callee object display throws [error here](https://github.com/kint-php/kint/blob/add-smarty-plugin/src/Object/TraceFrameObject.php#L95) if I do `parser->haltParse()` without hardcoding a `classname` member onto `BasicObject &$o` [manually](https://github.com/kint-php/kint/compare/add-smarty-plugin?expand=1#diff-68cf966c5e12593677ac4a4972935bddR93). If I don't `haltParse` - no error, but all Smarty dirt is visible again :)

I hope I'm explaining myself clearly, it's 3:30 in the morning here :) I also hope merging a pull request allows you to squash my commits into one.

Anyway, looking forward to hearing for you, and once again - great job :)

### DIFF
--- a/src/Kint.php
+++ b/src/Kint.php
@@ -154,7 +154,6 @@ class Kint
         'Kint\\Parser\\JsonPlugin',
         'Kint\\Parser\\MicrotimePlugin',
         'Kint\\Parser\\SimpleXMLElementPlugin',
-        'Kint\\Parser\\SmartyPlugin',
         'Kint\\Parser\\SplFileInfoPlugin',
         'Kint\\Parser\\SplObjectStoragePlugin',
         'Kint\\Parser\\StreamPlugin',
--- a/src/Kint.php
+++ b/src/Kint.php
@@ -154,6 +154,7 @@ class Kint
         'Kint\\Parser\\JsonPlugin',
         'Kint\\Parser\\MicrotimePlugin',
         'Kint\\Parser\\SimpleXMLElementPlugin',
+        'Kint\\Parser\\SmartyPlugin',
         'Kint\\Parser\\SplFileInfoPlugin',
         'Kint\\Parser\\SplObjectStoragePlugin',
         'Kint\\Parser\\StreamPlugin',
--- a/src/Parser/SmartyPlugin.php
+++ b/src/Parser/SmartyPlugin.php
@@ -1,5 +1,30 @@
 <?php
 
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2013 Jonathan Vollebregt (jnvsor@gmail.com), Rokas Šleinius (raveren@gmail.com)
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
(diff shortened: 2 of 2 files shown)


jnvsor:
Hi @raveren - I want to keep only core PHP related plugins in the main repo, which is why I made https://github.com/kint-php/kint-twig for a twig plugin and put it in `suggest`. Could we do the same for this? Perhaps we should make another repo `kint-extras` or something that just requires a bunch of useful plugins together.

> calling its `__toString` would cause exceptions

For the first issue with toString causing exceptions, I'd suggest using `ToStringPlugin::$blacklist` - seems like the simplest way to get it to stop.

> I did have some counter-intuitive problems to solve - precisely how to display a sub-element of a variable instead of the variable itself; this ended up working
> `$o = $this->parser->parse($var->value, $o);`
> maybe I should add this as a tip to the docs?

The parse method is demonstrated in the first code block on that page, and since `$o` is usually taken by reference, you should probably use a different variable or it should be overwriting the input object. (Which is what you want in your code but not in most cases where you're parsing sub info)

> Trace callee object display throws error here if I do parser->haltParse() without hardcoding a classname member onto BasicObject &$o manually. If I don't haltParse - no error, but all Smarty dirt is visible again :)

This is because you're running at `Parser::TRIGGER_BEGIN` - the base parser does *absolutely nothing* at this point, so you're required to wire up all the default stuff yourself (Example in [BlacklistPlugin](https://github.com/kint-php/kint/blob/add-smarty-plugin/src/Parser/BlacklistPlugin.php#L104)) I should add a note about that on the writing plugins page.

I haven't looked at the actual code yet - I'll give it a once-over when I get back. Do you want to make a new repo and continue this there or should we keep it in this PR until it's ready?

raveren:
Hey, can you make me maintainer of https://packagist.org/ kint-php vendor namespace, I can't add the plugin.

raveren:
All valid points, by the way, sorry for being brief.

jnvsor:
> can you make me maintainer of https://packagist.org/ kint-php vendor namespace

I can make you maintainers of individual packages, but I didn't even know there was a namespace thing. How do I do that? Is it already done now that I set you as maintainer?

Anyway, should I close this PR and write my comments on the new repo?

**Edit:** Went ahead and wrote it

jnvsor:
By the way @raveren did you want to add it to suggests?

raveren:
there's no need, googling "kint smarty" brings up the composer repo as first result, someday when there are more integrations they could be added to the readme.

PR #306, opened 2019-03-09
Rollbacked the removal of ddd

raveren:
ddd is back to dump and die!

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,10 @@ s($GLOBALS); // Basic output mode
 
 Kint::$enabled_mode = false; // Disable kint
 d('Get off my lawn!'); // Debugs no longer have any effect
+
+Kint::$enabled_mode = true; // Re-enable kint for the next demo
+ddd('this is the last you see of me!'); // ddd() and sd() output and die();
+echo 'This will not appear on page';
 ```
 
 ## Tips & Tricks
--- a/init_helpers.php
+++ b/init_helpers.php
@@ -80,10 +80,14 @@ if (!\function_exists('s')) {
         return $out;
     }
 
-    function sd(...$vars)
+    /**
+     * Kint::dump() plain mode and die();
+     */
+    function sd()
     {
-        if (!Kint::$enabled_mode) return 0;
-        s(...$vars);
+        if (!Kint::$enabled_mode) return;
+        $args = \func_get_args();
+        s($args);
         die;
     }
 
@@ -94,12 +98,12 @@ if (!\function_exists('s')) {
 if (!function_exists('ddd')) {
     /**
      * Kint::dump() and die();
-     * @param array $vars
      */
-    function ddd(...$vars)
+    function ddd()
     {
-        if (!Kint::$enabled_mode) return 0;
-        Kint::dump(...$vars);
+        if (!Kint::$enabled_mode) return;
+        $args = \func_get_args();
+        Kint::dump($args);
         exit;
     }
     
--- a/init_helpers.php
+++ b/init_helpers.php
@@ -82,3 +82,20 @@ if (!\function_exists('s')) {
 
     Kint::$aliases[] = 's';
 }
+
+
+
+if (!function_exists('ddd')) {
+    /**
+     * Kint::dump() and die();
+     * @param array $vars
+     */
+    function ddd(...$vars)
+    {
+        if (!Kint::$enabled_mode) return;
+        Kint::dump(...$vars);
+        exit;
+    }
+    
+    Kint::$aliases[] = 'ddd';
+}
--- a/init_helpers.php
+++ b/init_helpers.php
@@ -80,10 +80,14 @@ if (!\function_exists('s')) {
         return $out;
     }
 
-    Kint::$aliases[] = 's';
+    function sd(...$vars)
+    {
+        s(...$vars);
+        die;
+    }
+
+    Kint::$aliases[] = 'sd';
 }
-
-
 
 if (!function_exists('ddd')) {
     /**
--- a/init_helpers.php
+++ b/init_helpers.php
@@ -82,6 +82,7 @@ if (!\function_exists('s')) {
 
     function sd(...$vars)
     {
+        if (!Kint::$enabled_mode) return 0;
         s(...$vars);
         die;
     }
@@ -97,7 +98,7 @@ if (!function_exists('ddd')) {
      */
     function ddd(...$vars)
     {
-        if (!Kint::$enabled_mode) return;
+        if (!Kint::$enabled_mode) return 0;
         Kint::dump(...$vars);
         exit;
     }
--- a/init_helpers.php
+++ b/init_helpers.php
@@ -86,6 +86,7 @@ if (!\function_exists('s')) {
         die;
     }
 
+    Kint::$aliases[] = 's';
     Kint::$aliases[] = 'sd';
 }
 
--- a/init_helpers.php
+++ b/init_helpers.php
@@ -85,9 +85,17 @@ if (!\function_exists('s')) {
      */
     function sd()
     {
-        if (!Kint::$enabled_mode) return;
+        if (!Kint::$enabled_mode) return 0;
+
+        if (Kint::$enabled_mode !== Kint::MODE_TEXT) {
+            Kint::$enabled_mode = Kint::MODE_PLAIN;
+            if (PHP_SAPI === 'cli' && true === Kint::$cli_detection) {
+                Kint::$enabled_mode = Kint::$mode_default_cli;
+            }
+        }
+
         $args = \func_get_args();
-        s($args);
+        \call_user_func_array(array('Kint', 'dump'), $args);
         die;
     }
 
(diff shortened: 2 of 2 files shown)


jnvsor:
There were some complaints about `ddd` when it was first removed, but there haven't been for ages (Until you came back and relived it) I'd like to ask you to try it for a few weeks and see how it goes before putting this stuff back. (Personally, I've gotten used to just typing `exit` when I need it.) Here's some of my reasoning.

When I read your other issue I was going to bring up `sd` as well. How many helper functions do we need to make? Are we just winging it here?

In a 'perfect code' world we wouldn't define any global functions but that's annoying if you ever use namespaces. (Also more typing) I *would* like to keep them to a minimum if possible, so we should think carefully about what we actually need.

We need `d` because `\Kint::dump` is long and frequently used, and we need `s` because `\Kint::$enabled_mode=\Kint::MODE_PLAIN;d(...` is even longer and we have no modifier or automation for it like CLI and TEXT.

Making too many helper functions multiplies:

* `d` for dumping (1 helper)
* `s` for dumping plain (2 helpers)
* `ddd` and `sd` for dumping and exiting (4 helpers)
* If we wanted to add new simple on/off option we'd need 8 helper functions now

At some point while writing 2.x I had more than 6 and decided it was going the wrong direction, so we *do* need to think about this or it will cause problems.

On a more subtle note, these functions are untestable. If you `exit` in code the test suite will die too, so we would have to mark the helper functions as entirely `@codeCoverageIgnore` and hope someone notices manually if something breaks. ~~(Not that they're particularly complex and need testing)~~

**Edit:** After looking at the code I stand corrected. `sd` will `die` when kint is disabled and `ddd` will warn if kint is disabled and you use the `~` modifier on PHP 7+. (Ran into that myself a while back, and there are tests for it in `IntegrationTest.php`)

raveren:
Did my fixes address the issues in the edit? 

This is not philosophy, man, this is my workflow and that of thousands of others, you came and you disabled it, and that was rude and uncalled for, of course you got complaints from an otherwise quiet community.

The die functions stay, there has been too much discussion about it already.

jnvsor:
> this is my workflow and that of thousands of others

The others clearly got over it since the versions of Kint without `ddd` now have more than 10 times the daily downloads of the versions with it.

> you came and you disabled it, and that was rude and uncalled for

It was a major version upgrade. If I'm not allowed to push breaking changes in major version upgrades then when am I? Should I revert the last 170 commits to bring back PHP 5.1 support?

It's not like it wasn't on my master for months before you made the org, and it's not like it wasn't on kint-php/kint master and alpha/beta/rc tags for more than a month before it released, and it's not like I wasn't actively asking for feedback the whole time...

I did more than my due dilligence here.

> The die functions stay, there has been too much discussion about it already.

Stop using the word "stay". The last commit with `ddd` was 2016. This hasn't been in Kint releases for over 2 years. It hasn't been on master since the org was created. The last comment on the subject besides you was October 2017. The discussion was settled a long time ago, you just never showed up.

You are not rolling back a recent change, you are adding new code. You are the one making a change now.

I can't accept adding new global functions just because things changed in the years since you were last using Kint. Everyone else has either made their own `dd` helper or adapted to type `die;` themselves.

> Did my fixes address the issues in the edit?

And just to prove that this isn't some trivial code that doesn't need tests: You have a parse error on every PHP version under 5.6, you should allow `sd` to be initialized if `s` is already taken, and a closer look pointed out a tiny bug in the existing `s` and `Kint` that will have to wait because they're technically a BC break.

---

Sorry about the tone of this message, but I took some flak (Less than you seem to think I took) for getting rid of some code that needed to go, and I can't agree to let you put it back without at least having a proper reason other than "That's the way it was 3 years ago when I last used it".

I have to ask again: Please try to get used to it. If you can come up with a compelling argument to add these functions please do tell me, because I can't think of any pros that outweigh the cons.

raveren:
Every other dumper out there has the dump and die functionality, man, why do you think? 

Or if you want an answer, just read why do people opt to use the v1.1 version in new code: https://github.com/KnowTheCode/kint-php-debugger 

**EDIT:** this is not the link where people complain about there no longer being a `dd`, but I will not be looking for more proof of the obvious.

I mean why not, as far as users are concerned, 1.1 is virtually the same as `HEAD` functionality-wise, I *think* it got some minor additions, maybe bugfixes too, but had at least three regressions I mentioned elsewhere (plus the ill-advised `__toString` display).

I mean the work you've been doing is great, it's useful and valuable, but I will not be discussing the `ddd` issue anymore. As soon as I'm sure there is no bugs I will be merging this.

RE: `sd`: 
if `s` cannot be declared, `sd` can be not declared as well as far as I'm concerned, not having `s` is *already* something the user has to fix if he wants it.

jnvsor:
> Or if you want an answer, just read why do people opt to use the v1.1 version in new code: https://github.com/KnowTheCode/kint-php-debugger

Ok, let's have a read...

> Why not the latest Kint version?
> * Rendering is inconsistent with the video library on Know the Code - the new version wraps all of the dumps and packages them into one container. The container is rendered fixed at the bottom of the screen.
> * Composer required - the new version requires Composer to load all the files, which are extensive. To keep it simple, this plugin does not require Composer to run the plugin.

1. Turning off the folder is a one-liner
2. You can just use the phar file
3. **His reasons have nothing to do with `ddd`**

Do you have any other 4 day old repos with reasons to use 1.1 that are incorrect and have nothing to do with `ddd` or are you done now?

> **EDIT:** this is not the link where people complain about there no longer being a dd, but I will not be looking for more proof of the obvious.

I see.

> I mean why not, as far as users are concerned, 1.1 is virtually the same as HEAD functionality-wise, I think it got some minor additions, maybe bugfixes too, but had at least three regressions I mentioned elsewhere (plus the ill-advised __toString display).

Ah yes. Minor additions like access paths, search, and the ability to extend it without being forced to drop code inside kint's folders.

Little bugfixes like the iffy string encoding detection, recursion detection, call parsing, table display, and so on.

And three regressions so *huge* 2 of them are literally *documented configuration options*.

It's only 300 more commits than you've made in Kint's entire history, there can't be much in there right?...

> I will not be discussing the ddd issue anymore. As soon as I'm sure there is no bugs I will be merging this.

Earlier you called me rude for making a breaking change in a major version update, where breaking changes belong.

You come back after 2 and a half years of having no interaction with Kint besides the occasional thumbs up and start unilaterally inserting bad code despite my protests, into a project that [I am the lead developer on](https://github.com/kint-php/kint/graphs/contributors).

This issue is nothing if not rude.

Well if you insist on forcing an issue that only you have and pissing me off in the process, then you'll have no problem handling *any and all* issues coming from it.

I expect you to find a way to test it properly, too. I don't have 100% test coverage everywhere, but I damn well do on important stuff like that.

If you cause a problem and don't fix it in short order I'll remove it again, backwards compatibility be damned. And *when* that happens I'll forward all complaints to you.

raveren:
Hey, don't get angry, this is not about anything more than a dump and die functionality :)

It's an essential part of everyday workflow for many people, it must be present.

jnvsor:
Ignoring the question of how many people want to use it, I still think we need to discuss when we're going to add helper functions, or they will get out of hand.

For example, `sd` may be completely unnecessary - you might argue you would want a plain text dump for ajax calls and the like, but then you should probably be using `~dd`. In an html context `dd` should always be superior to `sd` unless you have already output some breaking html.

With that in mind you might also want to consider making `dd` an automatic `-dd` as well, to ensure it's the only thing dumped. Then there's the question of what you want to call it: `ddd` is likely to run into the same conflict problems `dd` did, but I have no solutions for this...

Regarding tests, I think the only way to properly test this would be with a separate script run as a subprocess. I wouldn't be opposed to making symfony/process a dev dependency for this case.

**Edit:** You might be able to hijack the `@runInSeparateProcess` annotation but I haven't looked into that

raveren:
Hey, I have been using this change, and it seems to not have any bugs, unless you have any real reason not to - I am going to merge this PR, say, tomorrow.

jnvsor:
Can you not see the big red "All checks have failed" at the bottom of the PR? Is there a setting somewhere I missed?

raveren:
ok yeah, travis is broken, so what do I do now, there are no error descriptions or anything

jnvsor:
> so what do I do now, there are no error descriptions or anything

Can you see the info if you click on details?

I have travis set up to do some small checks before going into a full multi-version test suite, notably:

* A single run of the tests on composer
* A `composer format` (Write all the sloppy code you want)
* A `composer build` (To ensure the phar file is up to date)
* A `composer analyze` (This thing catches loads of stuff)

Then it runs the test suite on both the composer and the phar file on all php versions to ensure everything's working.

Travis' nightly build is missing GD and broke a few tests so I have it set to allow_failure for now. Will need to tackle that properly before 7.4 releases.

Currently due to a problem with a dependency of a dependency, `composer format` doesn't work on 7.3 so you have to run it in 7.2. I have a patch in the pipeline to fix that though.

PR #307, opened 2019-03-09
Made the kint display wrapper at the bottom opt-in

raveren:
To turn it back on again use
`\Kint\Renderer\RichRenderer::$folder = true;`

this change should maybe be reflected in documentation..?

### DIFF
--- a/src/Renderer/RichRenderer.php
+++ b/src/Renderer/RichRenderer.php
@@ -118,7 +118,7 @@ class RichRenderer extends Renderer
      *
      * @var bool
      */
-    public static $folder = true;
+    public static $folder = false;
 
     /**
      * Sort mode for object properties.


jnvsor:
I'd rather not make changes to the default settings without a major version bump. People only complain when things change, after all.

I'd like a way to poll users or something, but given how people responded (Or rather, didn't respond) to my public calls for feedback on new major versions before, I feel it would be a waste of time.

Like #306 could you give this a try for a bit first? I added it because I kept running into situations where it would have been useful after all.

> this change should maybe be reflected in documentation

The option as-is is already documented on the site and in the readme, so it would be a case of flipping the meaning.

raveren:
Agreed about the postponing the change till a major version.

jnvsor:
Now that I look back at the issues, I can't find anyone actually complaining about it. I found a few people looking for documentation though. (Before I wrote said documentation)

Let's just leave this for now and revisit it when we're approaching a new major

PR #310, opened 2019-03-20
Make file path plugin work on windows

raveren:


jnvsor:
I'm confused, don't file paths on windows usually start with a drive number like `C:/...`, or does PHP have some cgywin-esque `/c/...` paths instead?

raveren:
no, this regex invalidates strings that have a `:` in them, which paths on windows always do.

jnvsor:
Oh, my bad!

I'm writing a comment on #306 about CI so you can get this through, but please hold off on pushing #306 for now

PR #335, opened 2020-05-19
Add rendering for spl_object_id

xelax90:


jnvsor:
All right, this is a good idea and you've put a lot of legwork into it (Kudos!) so I'd like to see this merged, so let me go through a few issues.

I can't do a proper github code review because it's based off my unfinished stuff that hasn't been pushed to kint-php yet so there's a bunch of extra commits in there, so I'll just make some bullet points

* `spl_object_id` is PHP 7.2+. Even in the next version of kint I intend to target 5.6+ so it can't be on by default, it would have to be behind version flags.
* Actually since I already have `spl_object_hash` I'd go a step further and say the rendering half should be an opt-in plugin for rendering both the hash and id in some sort of foldout like the access path button
* `identifier` is a bit vague - should probably be `spl_object_id` (But that's my bad for calling `spl_object_hash` `hash`) -- Maybe start with a commit renaming hash? Actually maybe identifier is a better name, I should check the zend source for the actual names used...  
  **Edit:** Never mind, the internal field name is `handle` and the hash is just a hash of the handle plus some random stuff - `spl_object_id` is much clearer
* The identifier should be part of `InstanceObject` like `hash`, strings, ints etc. don't have object ids
* Could use a test in `ParserTest` checking that the ID actually matches `spl_object_id`, but again needs a version flag

For now could you add a test to ParserTest and a version flag in Parser, plus move identifier to InstanceObject? We can deal with the nitpicking and renderer side of it once we have the parsing done

:+1:

xelax90:
I actually already added a test to ParserTest already. Should I maybe move my changes to a new branch? I kind of messed up and pushed the changes to my master and then made pull request to 3.x. I should propably just rebase my change onto 3.x, right?

jnvsor:
> I should propably just rebase my change onto 3.x, right?

Oh *that's* what happened. No just leave it for now - I have one more change I want to make to my master before a push to kint-php and then at least kint-php's master will be cleaned up and you can rebase onto that. (Though it's going to cause some conflicts - sorry in advance!)

The ParserTest looks good, could you put an if around the identifier assignment in parser using `KINT_PHP72` and move the identifier into InstanceObject? I'll try to get my stuff done this weekend and then I'll help you finish up

raveren:
You guys rock! Like, really sincerely and objectively I find Kint the most informative and easy to use variable dumper out there, it's a hidden gem - thanks for keeping it going!

- a happy user, raveren :)

jnvsor:
@xelax90 I've finished my changes and pushed them to master. Like I said there's a fairly large breaking change in that the entire `Kint\Object\XObject` thing has been renamed `Kint\Zval\XValue`.

I'll be renaming the hash property on `InstanceValue` but it shouldn't collide with your work so go ahead and rebase on `kint-php/master` when you're ready

jnvsor:
Closing since this PR has died

PR #339, opened 2020-07-21
Update README.md

userlond:
Fix the "Download the file" link


jnvsor:
I'm a little concerned with CI not validating this commit (Seems like something broke recently) but I'll merge this when I fix it, thanks!

jnvsor:
Merged. CI failed because it's such an old branch, but it's just a changed link so it's fine. Thanks for the fix!

Issue #203, opened 2016-05-21
Recursion bug - leaving marker in arrays in certain conditions


jnvsor:
Don't know why exactly but:

``` php
$a = ["Array 1"];
$b = ["Array 2"];
$a[] = &$b;
$b[] = &$a;

d($a);
d($b);
```

This shows the marker in $a in the second call.

I'm doing an experimental overhaul of the parser (It's looking good) so I'll probably come across more of these things.

raveren:
I'm sorry, I don't understand the issue, if the marker is there, this might just be something PHP does weirdly, no surprise there :) Kint just represents what data PHP holds internally...

jnvsor:
No, it's the marker kint put there to detect recursion - the marker at `kintParser::$_marker`

raveren:
Oh god you found the polterbug I've been looking to reproduce for for ages! Thank you, I fixed it, but it will come with a batch of changes later this week!

The really, terribly, scary hard task I have for the past two years is to bring documentation up to date :) I really want to include a _notable contributors_ section to give excellent developers like you due credit and immense thanks for all of your hard work!

jnvsor:
A more scary note on recursion handling is that kint is doing something that's theoretically impossible to do correctly. For example:

``` php
<?php

$a = ["Array 1"];
$b = ["Array 2"];
$a[] = &$b;
$b[] = &$a;
$c = ["Array 1", &$b];

d($a);
d($b);
d($c);
```

Clearly `$c` is not referenced by `$b`, but since kint gets it's arguments through `func_get_args` (By copy not reference) there is no way to correctly determine recursion in this case.

I think kint should (Like `print_r`, `var_dump` and other debug builtins which all take parameters by reference) stop trying to be clever.

To demonstrate why the "Clever" behaviour is probably a bad thing, it didn't take long to find a version that's full on completely wrong output:

``` php
<?php
$a = ["Array 1"];
$b = ["Array 2"];
$a[] = &$b;
$b[] = &$a;
$c = [&$b];

d($a);
d($b);
d($c);
```

raveren:
There is definitely an edge-case bug in detecting circular references, but it has nothing to do with the way Kint gets its arguments. It's a side effect of a conscious effort to short-circuit 

```
array (2) [
    string (7) "Array 1"
    array (2) [
        string (7) "Array 2"
        array (2) [
            string (7) "Array 1"
            *RECURSION*
        ]
    ]
]
```

into

```
array (2) [
[...]

PR #343, opened 2020-08-12
Recursion detection by reference 

kktsvetkov:
As discussed in https://github.com/kint-php/kint/issues/203 this is recursion tracking by reference. The array and object references are collected in a "register" and if they are encountered a second time, that is considered a recursion.

One of the benefits of this approach is that arrays are never tainted with markers, and at any time a plugin can pick the parsed array variable as it is.

I do use hashes to address the object references, but that's because it is faster that way instead of traversing the whole list of knownObjects.

@jnvsor has some concerns about memory leaks and performance, and I'll be looking forward for ways how to test these cases. So far I've only done the examples from https://github.com/kint-php/kint/issues/203 and the `ParserTest.php` unit test, and both work OK.

Structurally, the output of the recursions is the same as from `var_dump()`. Here's what I used to compare:
```php 
<?php

include 'vendor/autoload.php';

$a = ["Array 1"];
$b = ["Array 2"];
$a[] = &$b;
$b[] = &$a;

!d($a);
!d($b);

var_dump($a);
var_dump($b);

$r = (object) ['a' => 23, 'b' => 45];
$r->c =& $r;
!d($r);

var_dump($r);
```
There is a screenshot of the output [here](https://github.com/kint-php/kint/issues/203#issuecomment-673141991)

### DIFF
--- a/src/Parser/Recursion.php
+++ b/src/Parser/Recursion.php
@@ -1,0 +1,91 @@
+<?php
+
+namespace Kint\Parser;
+
+/**
+ * Recursion tracker
+ *
+ * Collects references to arrays and objects introduced during the
+ * inspection of a variable inside the {@link Kint\Parser\Parser}.
+ * Any variables introduced more than once constitutes recursion.
+ *
+ * @author Kaloyan KT Tsvetkov <kaloyan@gmail.com>
+ */
+final class Recursion
+{
+    /**
+     * @var array list of referenced arrays
+     * @see Recursion::isArrayRecursion();
+     */
+    protected $knownArrays = [];
+
+    /**
+     * Checks for array recursion
+     *
+     * @param array &$var
+     * @return bool
+     */
+    function isArrayRecursion(array &$var) : bool
+    {
+        foreach ($this->knownArrays as &$known)
+        {
+            if ($known === $var)
+            {
+                return true;
+            }
+        }
+
+        $this->knownArrays[] =& $var;
+        return false;
+    }
+
+    /**
+     * Cleans the collected array references
+     */
+    function cleanArrays()
+    {
+        $this->knownArrays = [];
+    }
+
+    /**
+     * @var array list of referenced objects
+     * @see Recursion::isObjectRecursion();
+     */
+    protected $knownObjects = [];
+
+    /**
+     * Checks for object recursion
+     *
+     * @param object &$var
+     * @return bool
+     */
+    function isObjectRecursion(&$var) : bool
+    {
+        // can not type-hint generic object values,
+	// so the best option is to use is_object()
+        if (!is_object($var))
+        {
+            return false;
+        }
+
+	// we can foreach the list of $knownObjects
+	// as we do with the arrays, but using the
+	// spl_object_hash() is faster
+        $hash = \spl_object_hash($var);
+        if (!empty($this->knownObjects[ $hash ]))
+        {
+            return true;
+        }
+
+        $this->knownObjects[ $hash ] =& $var;
+        return false;
+    }
+
+    /**
+     * Cleans the collected object references
+     */
+    function cleanObjects()
+    {
+        $this->knownObjects = [];
+    }
+}
--- a/src/Parser/Recursion.php
+++ b/src/Parser/Recursion.php
@@ -2,6 +2,11 @@
 
 namespace Kint\Parser;
 
+use function is_array;
+use function is_object;
+use function spl_object_hash;
+use function uniqid;
+
 /**
  * Recursion tracker
  *
@@ -20,6 +25,20 @@ final class Recursion
     protected $knownArrays = [];
 
     /**
+     * @var string marker for detecting array references vs copies
+     * @see Recursion::isArrayRecursion();
+     */
+    protected $marker;
+
+    /**
+     * Creates new {@link Kint\Parser\Recursion} object
+     */
+    function __construct()
+    {
+        $this->marker = uniqid("kint\0", true);
+    }
+
+    /**
      * Checks for array recursion
      *
      * @param array &$var
@@ -31,11 +50,27 @@ final class Recursion
         {
             if ($known === $var)
             {
-                return true;
+                // test if they are copies of one another,
+                // or they are reference to the same array
+                $var[ $this->marker ] = 1;
+                $recursion = !empty($known[ $this->marker ]);
+                unset($var[ $this->marker ]);
+                return $recursion;
             }
         }
 
-        $this->knownArrays[] =& $var;
+        // we are only interested in arrays with
+        // nested other arrays and objects inside,
+        // where the recursion can occur
+        foreach ($var as $k => &$v)
+        {
+            if (is_object($v) || is_array($v))
+            {
+                $this->knownArrays[] =& $var;
+                break;
+            }
+        }
+
         return false;
     }
 
@@ -71,7 +106,7 @@ final class Recursion
         // we can foreach the list of $knownObjects
         // as we do with the arrays, but using the
         // spl_object_hash() is faster
-        $hash = \spl_object_hash($var);
+        $hash = spl_object_hash($var);
         if (!empty($this->knownObjects[ $hash ]))
         {
             return true;
--- a/src/Parser/Recursion.php
+++ b/src/Parser/Recursion.php
@@ -55,7 +55,9 @@ final class Recursion
                 $var[ $this->marker ] = 1;
                 $recursion = !empty($known[ $this->marker ]);
                 unset($var[ $this->marker ]);
-                return $recursion;
+                if ($recursion) {
+                    return true;
+                }
             }
         }
 
--- a/src/Parser/Recursion.php
+++ b/src/Parser/Recursion.php
@@ -62,15 +62,15 @@ final class Recursion
     function isObjectRecursion(&$var) : bool
     {
         // can not type-hint generic object values,
-	// so the best option is to use is_object()
+        // so the best option is to use is_object()
         if (!is_object($var))
         {
             return false;
         }
 
-	// we can foreach the list of $knownObjects
-	// as we do with the arrays, but using the
-	// spl_object_hash() is faster
+        // we can foreach the list of $knownObjects
+        // as we do with the arrays, but using the
+        // spl_object_hash() is faster
         $hash = \spl_object_hash($var);
         if (!empty($this->knownObjects[ $hash ]))
         {
--- a/src/Parser/Recursion.php
+++ b/src/Parser/Recursion.php
@@ -46,33 +46,27 @@ final class Recursion
      */
     function isArrayRecursion(array &$var) : bool
     {
-        foreach ($this->knownArrays as &$known)
+        $count = count($var);
+        if (!empty($this->knownArrays[$count]))
         {
-            if ($known === $var)
+            $var[ $this->marker ] = 1;
+            $recursion = false;
+            foreach ($this->knownArrays[$count] as &$known)
             {
-                // test if they are copies of one another,
-                // or they are reference to the same array
-                $var[ $this->marker ] = 1;
-                $recursion = !empty($known[ $this->marker ]);
-                unset($var[ $this->marker ]);
-                if ($recursion) {
-                    return true;
+                if ($recursion = !empty($known[ $this->marker ]))
+                {
+                    break;
                 }
             }
-        }
 
-        // we are only interested in arrays with
-        // nested other arrays and objects inside,
-        // where the recursion can occur
-        foreach ($var as $k => &$v)
-        {
-            if (is_object($v) || is_array($v))
+            unset($var[ $this->marker ]);
+            if ($recursion)
             {
-                $this->knownArrays[] =& $var;
-                break;
+                return true;
             }
         }
 
+        $this->knownArrays[$count][] =& $var;
         return false;
     }
 
@@ -105,9 +99,6 @@ final class Recursion
             return false;
         }
 
-        // we can foreach the list of $knownObjects
-        // as we do with the arrays, but using the
-        // spl_object_hash() is faster
         $hash = spl_object_hash($var);
         if (!empty($this->knownObjects[ $hash ]))
         {
--- a/src/Parser/Parser.php
+++ b/src/Parser/Parser.php
@@ -27,6 +27,7 @@ namespace Kint\Parser;
 
 use DomainException;
 use Exception;
+use Kint\Parser\Recursion;
 use Kint\Zval\BlobValue;
 use Kint\Zval\InstanceValue;
 use Kint\Zval\Representation\Representation;
@@ -59,10 +60,10 @@ class Parser
 
     protected $caller_class;
     protected $depth_limit = false;
-    protected $marker;
-    protected $object_hashes = [];
     protected $parse_break = false;
     protected $plugins = [];
+
+    protected $recursion;
 
     /**
      * @param false|int   $depth_limit Maximum depth to parse data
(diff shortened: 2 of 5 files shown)


kktsvetkov:
To address the performance concerns, I've created a very crude benchmark test using the example from #203:
```php
<?php

include 'vendor/autoload.php';

$a = ["Array 1"];
$b = ["Array 2"];
$a[] = &$b;
$b[] = &$a;

$repeats = 9999;
d(microtime());
while ($repeats--)
{
@d($a);
@d($b);
}
d(microtime());
```
I am on High Siera and PHP 7.1
```
PHP 7.1.33 (cli) (built: Jan 18 2020 13:49:07) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.1.0, Copyright (c) 1998-2018 Zend Technologies
    with Xdebug v2.5.5, Copyright (c) 2002-2017, by Derick Rethans
```

Here's the benchmark output when running in [master](https://github.com/kktsvetkov/kint3/tree/master) branch:
```
┌───────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│ microtime()                                                                                                       │
└───────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
string (21) "0.13744700 1597272302"
    TIME: 2020-08-12 22:45:02.137447
    MEMORY USAGE: 1571048 bytes (1.5 MB) (real 4 MB).
    PEAK MEMORY USAGE: 1850136 bytes (1.8 MB) (real 4 MB).
═════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
Called from <ROOT>/benchmark.php:11 [d()]

┌───────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│ microtime()                                                                                                       │
└───────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
string (21) "0.79949700 1597272366"
    TIME: 2020-08-12 22:46:06.799497
    SINCE LAST CALL: 64.662s.
    SINCE START: 64.662s.
    AVERAGE DURATION: 64.662s.
    MEMORY USAGE: 2869800 bytes (2.7 MB) (real 6 MB).
    PEAK MEMORY USAGE: 6009968 bytes (5.7 MB) (real 6 MB).
═════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
Called from <ROOT>/benchmark.php:17 [d()]
```

Here's the benchmark output from [recursion_without_marker_and_hashes](https://github.com/kktsvetkov/kint3/tree/recursion_without_marker_and_hashes)
```
┌───────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│ microtime()                                                                                                       │
└───────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
string (21) "0.41824800 1597272775"
    TIME: 2020-08-12 22:52:55.418248
    MEMORY USAGE: 1579512 bytes (1.5 MB) (real 4 MB).
    PEAK MEMORY USAGE: 1850136 bytes (1.8 MB) (real 4 MB).
═════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
Called from <ROOT>/benchmark.php:11 [d()]

┌───────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│ microtime()                                                                                                       │
└───────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
string (21) "0.28937800 1597272826"
    TIME: 2020-08-12 22:53:46.289378
    SINCE LAST CALL: 50.8711s.
    SINCE START: 50.8711s.
    AVERAGE DURATION: 50.8711s.
    MEMORY USAGE: 3018264 bytes (2.9 MB) (real 6 MB).
    PEAK MEMORY USAGE: 6014080 bytes (5.7 MB) (real 6 MB).
═════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
Called from <ROOT>/benchmark.php:17 [d()]
```
Please run the benchmark on your end and let's compare results.

jnvsor:
> You are correct, === compares contents only and will produce false positives when arrays with the same contents are met again.

This is a pretty fatal flaw in my opinion. The current implementation handles this correctly. I'd appreciate it if you could add a test to distinguish this (In another PR so we can just merge it right away)

This version also *will* have massive performance issues with larger arrays. Instead of some small ones, consider benchmarking with very *deep* arrays and no depth limit, since it will be doing a deep comparison to check for equality, and then recursing in the parser and doing it again for each level until the depth limit.

This is O(n log n) compared to the current O(n) and should be much slower unless reads are *much* faster than writes in your system. (Never did learn big O complexity properly, I may be mistaken)

A way to do it with references might be to swap the variables out and see if it changed. Something like this:

```
function isArrayRecursion(array &$var) : bool
{
    foreach ($this->knownArrays as &$known)
    {
        $tmp = $known;
        $known = $this->marker;

        if ($var === $this->marker)
        {
            $known = $tmp;
            return true;
        }

        $known = $tmp;
    }

    $this->knownArrays[] =& $var;
    return false;
}
```

Since PHP uses COW variables that shouldn't create too much overhead even with large variables, but it's still O(n^2) because you have to loop over the list of known arrays for each level of depth.

I have a few other issues here but I'm not going to bother you with them until we know for sure if it works or not.

**Edit:** I saw your edit:

> Perhaps the arrays can be tainted with the markers only when such a match is found in order to detect if the two arrays are copies of one another, OR one is a reference of the other.

Yeah that's a sort of halfway between your version and mine. Go with that idea

> Also, I think that there can be fewer arrays tracked for recursion since if all the array elements are scalar (or resources), there is no way for recursion to occur: you actually need "nestable" data types such as arrays and objects to have a recursion.

I think the overhead of checking the types of the contents of a theoretically infinitely wide array is much higher than a superfluous write and delete in a hash table where those are O(1)

src/Parser/Recursion.php
```diff
@@ -31,11 +50,27 @@ function isArrayRecursion(array &$var) : bool
         {
             if ($known === $var)
             {
-                return true;
+                // test if they are copies of one another,
+                // or they are reference to the same array
+                $var[ $this->marker ] = 1;
+                $recursion = !empty($known[ $this->marker ]);
+                unset($var[ $this->marker ]);
+                return $recursion;
```
jnvsor: This will still return if `$known === $var` so it will provide false *negatives* now (Which is worse than false positives!)

kktsvetkov: @jnvsor nice catch, sorry about that

kktsvetkov: Changed it in 772415a and now the return is triggered only when there is a recursion detected

jnvsor:
So I wrote a benchmark to test a recursing 10k deep array to measure the parsing time difference:

```php
<?php

require __DIR__ . '/../path/to/kint/vendor/autoload.php';

ini_set('memory_limit', '2G');

$final = [];
$array = [&$final];

for ($i = 0; $i < 10000; $i++) {
    $array = [$array];
}

$final[] = $array;

var_dump($array);

Kint::$max_depth = 0;
$statics = Kint::getStatics();
$kintstance = Kint::createFromStatics($statics);
$kintstance->setStatesFromStatics($statics);
$parser = $kintstance->getParser();

$v = Kint\Zval\Value::blank('$array', '$array');

$t1 = microtime(true);
$x = $parser->parse($array, $v);
$t2 = microtime(true);

var_dump($t2 - $t1);

var_dump(memory_get_peak_usage());
```

Unfortunately, your implementation seems to have an infinite recursion somewhere... That or it's not correctly handling the recursion in the arrays.

> PHP Fatal error:  Nesting level too deep - recursive dependency? in .../kint/src/Parser/Recursion.php on line 51

kktsvetkov:
Thanks for the example. I'll have a look over the weekend to see what's going on.

kktsvetkov:
I haven't had the chance to do any testing yet, but just by reading this, I think the flaw in my code is caused by the depth-first traversing. The first-ever element of the array is the actual reference to itself, so there is never anything to compare against, so it goes deeper and deeper. I only ever tested where the recursion was the last element of the array, which now seems extremely stupid :)

kktsvetkov:
Ahm @jnvsor have you ran that benchmark script against master branch of Kint ? I just did and I got the similar output:

```
~/kt.kt/kint-php/🐱 php benchmark.php 
/Users/polina/kt.kt/kint-php/benchmark.php:16:
array(1) {
  [0] =>
  array(1) {
    [0] =>
    array(1) {
      [0] =>
      array(1) {
        ...
      }
    }
  }
}
PHP Fatal error:  Maximum function nesting level of '256' reached, aborting! in /Users/polina/kt.kt/kint-php/src/Parser/ArrayLimitPlugin.php on line 72
PHP Stack trace:
PHP   1. {main}() /Users/polina/kt.kt/kint-php/benchmark.php:0
PHP   2. Kint\Parser\Parser->parse() /Users/polina/kt.kt/kint-php/benchmark.php:27
PHP   3. Kint\Parser\Parser->parseArray() /Users/polina/kt.kt/kint-php/src/Parser/Parser.php:156
PHP   4. Kint\Parser\Parser->parse() /Users/polina/kt.kt/kint-php/src/Parser/Parser.php:405
...
PHP 254. Kint\Parser\Parser->parse() /Users/polina/kt.kt/kint-php/src/Parser/Parser.php:405
PHP 255. Kint\Parser\Parser->applyPlugins() /Users/polina/kt.kt/kint-php/src/Parser/Parser.php:150
PHP 256. Kint\Parser\ArrayLimitPlugin->parse() /Users/polina/kt.kt/kint-php/src/Parser/Parser.php:587
~/kt.kt/kint-php/🐱 git branch
* master
  recursion_without_marker_and_hashes
```
Can you check on your end as well just in case I am doing something wrong? The error is not in the parser though, it is at `src/Parser/ArrayLimitPlugin.php:72`. I removed that plugin (e.g. `unset(Kint::$plugins[0]);`, it's the first on the list), but then that happened again at a different place:
```
PHP Fatal error:  Maximum function nesting level of '256' reached, aborting! in /Users/polina/kt.kt/kint-php/src/Zval/Value.php on line 206
PHP Stack trace:
PHP   1. {main}() /Users/polina/kt.kt/kint-php/benchmark.php:0
PHP   2. Kint\Parser\Parser->parse() /Users/polina/kt.kt/kint-php/benchmark.php:29
...
```

jnvsor:
> Maximum function nesting level of '256' reached, aborting!

Yeah that's xdebug, you'll have to turn it off to get it to work or raise the limit: `xdebug.max_nesting_level=99999999999999`

(Incidentally with xdebug on it took much longer to run and used much more memory and I gave up after setting memory limit to 16G wasn't enough)

kktsvetkov:
I ended up lowering the nesting level instead to just 10
```php 
for ($i = 0; $i < 10; $i++) ...
```

kktsvetkov:
> PHP Fatal error: Nesting level too deep - recursive dependency? in .../kint/src/Parser/Recursion.php on line 51

The 51 line is `if ($known === $var)` and that seems to be a known problem when comparing arrays with circular references inside them because for this comparison (as well as just `==`) PHP digs deep to get every element value and compare them. 

To demonstrate it, here's a quick snippet:
```php 
$a = [];
$b = [];
$a[] =& $b;
$b[] =& $a;
var_dump($a === $b);

PHP Fatal error:  Nesting level too deep - recursive dependency? in php shell code on line 1
PHP Stack trace:
PHP   1. {main}() php shell code:0
```

The way to not have this happen is to not compare arrays :)

jnvsor:
Pretty sure that settles it then. No other way to check than by inserting a marker. I probably ran into the same thing when I was making 2.x.

I'm going to close this for now. If you have another way of handling it feel free to push more to this PR

kktsvetkov:
I'll try without comparison and tainting with markers outside only when inspecting.

Indeed using some sort of marking is recommended in several discussions I read online. I looked at php-src and it seems that there recursion is a separate type of variable. I still haven't tested var-dumper locally and how it handles these deep references

kktsvetkov:
In 7.4 there is \ReflectionReference, and in the RFC the `VarCloner` example for detecting references is exactly the "marker" approach :) 

> As an example, the VarCloner, which is part of Symfony's debug dump component uses essentially the following approach to detect references:
> ```php
> $array2 = $array1;
> $array2[$key] = $unique_cookie;
> if ($array1[$key] === $unique_cookie) {
>    // $array1[$key] is a reference
> }
> ```

https://wiki.php.net/rfc/reference_reflection

That is actually the same approach Kint has in detecting references inside array elements with `$refmarker`. If you would agree that recursions appear only there where are references, there is certainly some overlap with recursion detection.

jnvsor:
> If you would agree that recursions appear only there where are references

Well there are no pointers in php so yeah. Of course all objects are passed by reference too

kktsvetkov:
Yes, objects are passed by reference too, but it is the arrays that present the challenge. The objects hash approach is very good since it does not modify the object, or (as with the arrays) leave it stained with the marker inside.

That's why I am interested in the "arrays are references" matches, since this is where the recursion will manifest, and in this way, we can monitor a lot fewer arrays, and still add the marker only temporary just to do the detection.

kktsvetkov:
BTW, even with `ini_set('xdebug.max_nesting_level', '99999999999999')` the benchmark.php still crashes for me with `Segmentation fault: 11`
```
~/kt.kt/kint-php/🐱 php benchmark.php 
/Users/polina/kt.kt/kint-php/benchmark.php:17:
array(1) {
  [0] =>
  array(1) {
    [0] =>
    array(1) {
      [0] =>
      array(1) {
        ...
      }
    }
  }
}
Segmentation fault: 11
```
I settled on 1000 levels depth, with the `ArrayLimitPlugin` removed:
```
<?php

include 'vendor/autoload.php';

ini_set('memory_limit', '2G');
ini_set('xdebug.max_nesting_level', '99999999999999');

$final = [];
$array = [&$final];

for ($i = 0; $i < 1000; $i++) {
    $array = [$array];
}

$final[] = $array;

var_dump($array);

Kint::$max_depth = 0;
unset(Kint::$plugins[0]);	// disable ArrayLimitPlugin
$statics = Kint::getStatics();
$kintstance = Kint::createFromStatics($statics);
$kintstance->setStatesFromStatics($statics);
$parser = $kintstance->getParser();

$v = Kint\Zval\Value::blank('$array', '$array');

$t1 = microtime(true);
$x = $parser->parse($array, $v);
$t2 = microtime(true);

var_dump($t2 - $t1);
var_dump(memory_get_peak_usage());
```
For `recursion_without_marker_and_hashes` branch the output is:
```
/Users/polina/kt.kt/kint-php/benchmark.php:32:
double(0.28082990646362)
/Users/polina/kt.kt/kint-php/benchmark.php:33:
int(18554864)
```

For `master` branch the output is
```
/Users/polina/kt.kt/kint-php/benchmark.php:32:
double(0.29722905158997)
/Users/polina/kt.kt/kint-php/benchmark.php:33:
int(18543608)
```

I ran those several times in each branch, and more or less they are the same. The tiny benefit from my approach is that arrays are left without the marker inside them, which makes `getCleanArray()` useless.

kktsvetkov:
It seems since the PR is closed the new commit pushed to https://github.com/kktsvetkov/kint3/tree/recursion_without_marker_and_hashes is not being added here to the PR.

jnvsor:
Ok I'll open this again and take a look when I have some time

jnvsor:
Well with your latest commit I'm getting an out of memory at 2G in that test script, while normal kint finishes in 0.1ms using only 1 meg of ram. I'm not sure what's going on there but you might want to run with xdebug in profile mode and cachegrind it to see where it's going wrong.

All that said - even when finished this design is clearly doing more work than the current master, and there's not much point trying to do it with references when #203 has been fixed for quite a while.

kktsvetkov:
My goal with the reference research was to have the arrays not carrying the markers. I respect your concerns and without the tests showing any significant change in performance, let's indeed close this PR. Thanks for the time spent on this with me

jnvsor:
No problem. If you have any other ideas feel free to submit more PRs

Issue #349, opened 2020-10-02
Update docs on website


jnvsor:
Things that need checking in docs

* [x] Parser: Rename 'unknown' to 'resource (closed)' to align with PHP7.2+
* [x] Disable folder by default
* [x] Set minimum PHP version to 5.6 and update dependencies
* [x] Parser: Measure string size in bytes
* [x] Utils::getHumanReadableBytes: Fix rounding errors
* [x] CallFinderTest: Add check for short function definitions
* [x] TextRenderer: Whitelist ArrayObjectPlugin
* [x] Rename Object namespace Zval
* [x] ArrayLimitPlugin and ElidedObject stuff 94af838c02da206621ea1d014a018afaf1d31a57
* [x] Zval renaming
    * [x] BasicObject -> Value
    * [x] ElidedObject -> ElidedValues
    * [x] *Object -> *Value
* [x] Rename InstanceValue::$hash to $spl_object_hash
* [x] RichRenderer: Change ObjectPluginInterface to ValuePluginInterface
* [ ] Allow facade subclassing with LSB
* [x] Remove d(1) functionality
* [x] Separate keyboard control context for inside and outside folder
* [x] XMLPlugin::xmlToSimpleXML: Bugfix falsey return
* [x] Don't fail on uninitialized static properties 54edef1c3166ad700c02d595a43b82bb0c02682d
* [x] PHP8 support
* [x] Xml: Support namespaces in DOMDocument and SimpleXML
* [x] Parser: Remove parseDeep
* [x] Kint: Rename max_depth to depth_limit
* [x] Another composer update and update Rich/BinaryPlugin to fix complaints c65660570cf64f0905836045f5dcba2a9f696422
* [x] SimpleXMLElement: Check to see if it's a "falsey object" before parsing
* [x] Move from travis to GHA
* [x] Uninitialized properties
* [x] Utils?

PR #355, opened 2021-01-15
Update settings.md

raveren:
- Lifted `Kint\Renderer\RichRenderer::$folder` to be more prominent, since Kint, very unfortunately, still defaults to true. Lifted `Kint::$file_link_format` too since it's an extremely useful feature.
- Added useful settings example to `Kint::$file_link_format`
- Listed existing themes
- Added instructions on where to put settings in a modern framework and linked my personal settings.

### DIFF
--- a/settings.md
+++ b/settings.md
@@ -17,9 +17,26 @@ title: Kint settings
 
 # Kint settings
 
-These are global Kint settings. For plugin settings please see the [plugins page]({{ site.baseurl }}/plugins/).
+These are global Kint settings. For plugin settings please see the [plugins page]({{ site.baseurl }}/plugins/). 
 
 <section id="common" markdown="1">
+
+## Usage with a framework
+
+Chances are you're using `composer`, in that case the cleanest way is to add an entry to the `autoload.files` configuration:
+
+```json
+    "autoload": {
+        "psr-4": {
+            "App\\": "src/"
+        },
+        "files": [
+            "config/kint.php"
+        ]
+    },
+```
+
+Place your settings there (`"config/kint.php"`). [Here is what settings I personally use](https://gist.github.com/raveren/eba373d8abb572b0528c73d145103f95).
 
 ## Most common settings
 
--- a/settings.md
+++ b/settings.md
@@ -25,13 +25,13 @@ These are global Kint settings. For plugin settings please see the [plugins page
 
 These are the settings you're most likely going to want to customize Kint behavior
 
-`Kint::$enabled_mode` | Determines what mode Kint will run in. `true` is automatic. `false` is disabled. Any of the keys to `Kint::$renderers` to select a specific mode. You should disable Kint in production, or only enable it per `$_SERVER['REMOTE_ADDR']`. Default `true`
+`Kint::$enabled_mode` | Determines what mode Kint will run in. `true` is automatic. `false` is disabled. Any of the keys to `Kint::$renderers` to select a specific mode. You should disable Kint in production, or only enable it per `$_SERVER['REMOTE_ADDR']`. Default `true`.
+`Kint\Renderer\RichRenderer::$folder` | Whether to move all kint dumps into a folder attached to the bottom of the viewport. Default `true`.
 `Kint::$plugins` | List of enabled plugins. Either string class names or instances. Plugins added through the Kint class will be cached and need to work with multiple parser instances. See [plugins page]({{ site.baseurl }}/plugins/) for more information.
 `Kint::$max_depth` | The maximum depth to parse. 0 for unlimited. Tweak this to balance performance and verbosity. Default 6.
 `Kint::$aliases` | List of helper function aliases. Either string for a function name, or array of two strings for a static method.
-`Kint::$app_root_dirs` | An array of paths to aliases. These will be replaced in the mini trace, backtraces, etc. Default maps `$_SERVER['DOCUMENT_ROOT']` to `<ROOT>`
-`Kint\Renderer\RichRenderer::$theme` | Which theme to use. One of the CSS files from `resources/compiled/`, or the full path to a CSS file. Default `original.css`
-`Kint\Renderer\RichRenderer::$folder` | Whether to move all kint dumps into a folder attached to the bottom of the viewport. Default `true`
+`Kint::$app_root_dirs` | An array of paths to aliases. These will be replaced in the mini trace, backtraces, etc. Default maps `$_SERVER['DOCUMENT_ROOT']` to `<ROOT>`.
+`Kint\Renderer\RichRenderer::$theme` | Which theme to use. One of the CSS files from `resources/compiled/`: `original`, `aante-light`, `solarized`, `solarized-dark` or the full path to a CSS file. Default `original.css`.
 
 </section>
 <section id="kint" markdown="1">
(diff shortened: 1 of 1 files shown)


jnvsor:
Hey, I have a few suggestions, but I needed to make a lot of changes to the docs in prep for 4.0 anyway (#349) would you mind if I make this PR/branch the WIP for that issue? Then we can switch the docs when 4 goes live (Or beta)

I'll leave a review with a few things in a minute

settings.md
```diff
@@ -17,21 +17,38 @@ title: Kint settings
 
 # Kint settings
 
-These are global Kint settings. For plugin settings please see the [plugins page]({{ site.baseurl }}/plugins/).
+These are global Kint settings. For plugin settings please see the [plugins page]({{ site.baseurl }}/plugins/).
```
jnvsor, line 20: Trailing space

settings.md
```diff
@@ -17,21 +17,38 @@ title: Kint settings
 
 # Kint settings
 
-These are global Kint settings. For plugin settings please see the [plugins page]({{ site.baseurl }}/plugins/).
+These are global Kint settings. For plugin settings please see the [plugins page]({{ site.baseurl }}/plugins/). 
 
 <section id="common" markdown="1">
 
+## Usage with a framework
```
jnvsor, line 24: The composer instructions are on the front page of the site, and I don't want to recommend changing autoloader settings to include a file or mess with autoloading since most frameworks and custom code already have places dedicated to doing that sort of stuff. It's "Just PHP" so I don't think we should be opinionated on how to config it

settings.md
```diff
@@ -17,21 +17,38 @@ title: Kint settings
...
+        "psr-4": {
+            "App\\": "src/"
+        },
+        "files": [
+            "config/kint.php"
+        ]
+    },
+```
+
+Place your settings there (`"config/kint.php"`). [Here is what settings I personally use](https://gist.github.com/raveren/eba373d8abb572b0528c73d145103f95).
```
jnvsor, line 39: We should make a "Common suggestions" page for this. I know blacklisting DI containers is something you should almost always do for example

settings.md
```diff
@@ -17,21 +17,38 @@ title: Kint settings
...
+
 ## Most common settings
 
 These are the settings you're most likely going to want to customize Kint behavior
 
-`Kint::$enabled_mode` | Determines what mode Kint will run in. `true` is automatic. `false` is disabled. Any of the keys to `Kint::$renderers` to select a specific mode. You should disable Kint in production, or only enable it per `$_SERVER['REMOTE_ADDR']`. Default `true`
+`Kint::$enabled_mode` | Determines what mode Kint will run in. `true` is automatic. `false` is disabled. Any of the keys to `Kint::$renderers` to select a specific mode. You should disable Kint in production, or only enable it per `$_SERVER['REMOTE_ADDR']`. Default `true`.
+`Kint\Renderer\RichRenderer::$folder` | Whether to move all kint dumps into a folder attached to the bottom of the viewport. Default `true`.
```
jnvsor, line 46: This is `false` by default in master/4.x so we should leave it at the bottom of this section or move it out of "Most common settings" entirely

settings.md
```diff
@@ -17,21 +17,38 @@ title: Kint settings
...
 `Kint::$max_depth` | The maximum depth to parse. 0 for unlimited. Tweak this to balance performance and verbosity. Default 6.
 `Kint::$aliases` | List of helper function aliases. Either string for a function name, or array of two strings for a static method.
-`Kint::$app_root_dirs` | An array of paths to aliases. These will be replaced in the mini trace, backtraces, etc. Default maps `$_SERVER['DOCUMENT_ROOT']` to `<ROOT>`
-`Kint\Renderer\RichRenderer::$theme` | Which theme to use. One of the CSS files from `resources/compiled/`, or the full path to a CSS file. Default `original.css`
-`Kint\Renderer\RichRenderer::$folder` | Whether to move all kint dumps into a folder attached to the bottom of the viewport. Default `true`
+`Kint::$app_root_dirs` | An array of paths to aliases. These will be replaced in the mini trace, backtraces, etc. Default maps `$_SERVER['DOCUMENT_ROOT']` to `<ROOT>`.
+`Kint\Renderer\RichRenderer::$theme` | Which theme to use. One of the CSS files from `resources/compiled/`: `original`, `aante-light`, `solarized`, `solarized-dark` or the full path to a CSS file. Default `original.css`.
```
jnvsor, line 51: Could you add the `.css` to the end of the examples? If someone just pastes in `aante-light` for instance it will fail

settings.md
```diff
@@ -43,7 +60,7 @@ Other Kint settings
 `Kint::$cli_detection` | Whether to detect if it's being run in a CLI and adjust the renderer. Default `true`
 `Kint::$display_called_from` | Whether to display the called from (mini trace) information. Default `true`
 `Kint::$expanded` | Whether to expand objects by default. Default `false`
-`Kint::$file_link_format` | A format to link source code paths to. Default `ini_get('xdebug.file_link_format')`
+`Kint::$file_link_format` | A format to link source code paths to. Easiest way to open files in PhpStorm directly from Kint output is to install the PhpStorm [`Remote Call`](https://plugins.jetbrains.com/plugin/6027-remote-call) plugin from inside the IDE settings and set `Kint::$file_link_format = 'http://localhost:8091/?message=%f:%l';`. Default `ini_get('xdebug.file_link_format')`
```
jnvsor, line 63: Can we make this shorter? Something like:

> A format to link source code paths to. For example PhpStorm remote call linking: `http://localhost:8091/?message=%f:%l` Default `ini_get('xdebug.file_link_format')`

jnvsor (changes requested):

settings.md
raveren: Did you understand what this chapter is about? 

If you have a different approach to what is being suggested there, please propose your way. Or if I missed this information somewhere else, please point it to me, cause it's not on the front page of the site.

settings.md
raveren: I'm making a 80% improvement with 20% of the effort here, I don't know what does blacklisting the DI containers would do, either.

settings.md
raveren: when it is that way, I'll make sure to do this myself, until it is though - this default is the #1 reason I do not recommend Kint to people now.

settings.md
raveren: oh, ok, thanks, will do!

settings.md
raveren: I'll try to shorten it, thanks.

settings.md
jnvsor: > Usage with a framework

The problem is that every framework has a different way of doing it. Quite a few have different ways of doing it in a single framework (IE xml/yml/php in symfony if it's added as a flex component or something)

Since the config (currently) is just a bunch of static props, I think we shouldn't be opinionated on how to config it.

I especially think giving people a copy paste `autoloader` section is asking for trouble when we get issues complaining about how "Kint broke our system!" because their application namespace wasn't `App` or their framework has multiple namespaces in the root composer.json or something.

settings.md
jnvsor: DI containers tend to contain _every_ instance in your application (Think symfony/container) so if kint comes across it it'll end up trying to dump a _lot_ of stuff. Is a bit of a perf killer in modern applications, so you want to blacklist it, but even though it's used a lot it's not core PHP so I didn't add it to the docs or as a default. It would fit in a separate page though. (So would a `dd` reimplementation)

If there's something wrong with the default config we should fix it in the default config, if not it feels like "Extra"

In your gist you've got:

* Disable folder: Default in next
* PhpStorm: Definitely not core PHP. Sublime text uses `subl://open?url=file://%f&line=%l` for instance, and I don't want to add editor-specific settings.
* Helpers: I have nothing against adding helpers, but I want to keep them to a minimum in vanilla, and I don't want to write up extra docs for them either

I still think they'd be better under separate headings on a separate page, that way if we're going to offer bits of your-mileage-may-vary code then people can pick and choose which parts they want

PR #372, opened 2021-12-16
Revert regression: reintroduce dump and die via kd and sd.

raveren:


raveren:
just squash the commits if you merge this @jnvsor

PR #373, opened 2021-12-16
Revert regression: reintroduce dump and die via kd and sd.

raveren:


raveren:
I have no idea what the checks are checking but I can't run `composer build` on windows nor are there any differences produced with the `composer format`. 

Please just merge these changes and do a release asap.

PR #403, opened 2023-01-02
Fix #402

### LINKED ISSUE #402: PHP Warning: realpath(): open_basedir restriction in effect
Hi,

A warning occurs when using Kint on the CLI with the `open_basedir` setting in the `php.ini` file: 
```log
PHP Warning: realpath(): open_basedir restriction in effect. File(C:\Windows\System32) is not within the allowed path(s): (...) in my-project\vendor\kint-php\kint\init.php on line 57
```

This happens because of these lines:
https://github.com/kint-php/kint/blob/c70060777e7ee83946625673e72cae846becddfc/init.php#L53-L60

This is very easy to solve:
```diff
-    if (false !== \realpath($_SERVER['DOCUMENT_ROOT'])) {
+    if (is_dir($_SERVER['DOCUMENT_ROOT'])) {
```

ghost:
A quick fix to solve #402.


ghost:
Hi @jnvsor, i don't understand why the build check fails. I executed the `composer build` command. Could you help me with this please?

build.php
```diff
@@ -30,11 +30,12 @@
 
 require_once __DIR__.'/vendor/autoload.php';
 
-\mkdir(__DIR__.'/build');
+$outdir = __DIR__.'/build';
+\is_dir($outdir) || \mkdir($outdir);
```
ghost, line 34: To execute `mkdir` only if the directory does not exist. This prevents the warning below:
```log
Warning: mkdir(): File exists in kint\build.php on line 33
```

build.php
```diff
@@ -30,11 +30,12 @@
...
 
-$outpath = __DIR__.'/build/kint.phar';
+$outpath = "{$outdir}/kint.phar";
 
-\unlink($outpath);
+!\is_file($outpath) || \unlink($outpath);
```
ghost, line 38: To execute `unlink` only if the file exists. This prevents the warning below:
```log
Warning: unlink(kint/build/kint.phar): No such file or directory in kint\build.php on line 37
```

jnvsor:
I'm still getting this warning when using `is_dir` or `file_exists` outside of the open_basedir:

```
$ php -d open_basedir=/home -r 'var_dump(is_dir("/usr"));'
Cannot load Xdebug - it was already loaded
PHP Warning:  is_dir(): open_basedir restriction in effect. File(/usr) is not within the allowed path(s): (/home) in Command line code on line 1
PHP Stack trace:
PHP   1. {main}() Command line code:0
PHP   2. is_dir($filename = '/usr') Command line code:1
Command line code:1:
bool(false)
```

It doesn't look like there's any way to check this beforehand other than looking up the setting and checking the path every time, which would be enormously annoying and performance intensive considering it happens in traces and filepath dumps too... But according to the docs:

>  All symbolic links are resolved, so it's not possible to avoid this restriction with a symlink.

So we couldn't even look up the setting and compare the paths, we'd have to check the entire path from the root up to make sure none of them are symlinks... And we can't do that since those symlinks may be outside the open_basedir! (Or in other words, we can't tell if something's in the open_basedir without the `realpath` call that causes the warning)

Nice little catch-22 so this is technically impossible to fix, unless you know of some function I don't see mentioned in the docs that can detect whether something is accessible despite open_basedir

> Hi @jnvsor, i don't understand why the build check fails. I executed the `composer build` command.

After the build you should have a changed `kint.phar` file to commit

jnvsor:
Well if we're using `@` anyway then we could just use the original `realpath` code. Using error-suppression is considered "bad code" and when php screaming is enabled it still won't stop the warnings.

The only place we use it right now is in `MysqliPlugin` where we have no choice, and it's both noted in the documentation and the reason why that plugin is opt-in (Otherwise it'd just be on by default)

I did take a moment to survey the code and found 3 other places it's used. Another opt-in plugin, legacy encoding detection (I'll add them to the docs) and another place where it looks like it's a leftover from a loooong time ago and should probably be removed.

> This is exactly what i did.

Might be a CRLF issue, since git presumes the phar is binary but the build script is text, building with CRLF would alter the phar stub's newlines... In which case it's a problem with your windows git setup

**Edit:** After further digging it looks like PHP screaming may have been removed... If that's the case I'd be ok with adding one here for this one edge case, but I need to make sure first

**Edit 2:** Looks like it's worse than before:

> If you have set your own default error handler then the error still gets sent to the error handler regardless of the @ sign.

That means we _really_ don't want any warnings regardless of whether we use `@` or not... You know how frameworks like to spin up whoops and stuff like that at the first sign of a notice

jnvsor:
> You're wrong. What you're describing is an old behavior.

https://3v4l.org/i8oYo - Something's wrong with your system, because in all PHP versions 3v4l hit it called the error handler all 3 times, though 2 of them used `@`.

Notably in PHP 8 the fake errno variable passed to the callback changed but it's definitely calling the handler even while suppressed

We can push a dummy error handler and pop it afterwards to achieve the same effect as `@` without triggering custom ones, but that also feels quite dirty and I imagine the overhead would be massive too... Not sure what the way to go here is

**Edit:** I see your point, the warning _isn't_ sent even though the custom error handler _is_ called but that's my point. Custom error handlers from frameworks usually stop the world for every warning and throw up a detailed backtrace. Depending on the handler it may ignore the ones that have been suppressed but it may not (It's up to the handler)

I'll dig into symfony's error handler and whoops to see what they do with errno. If they pay attention to it then I guess we can rely on `@` most of the time

jnvsor:
Whoops respects the suppression by default, while symfony's error handler has a special option to scream specific error levels (But not warnings or notices by default)

So basically we can use `@` with little negative effect now.

Are you sure the one in the init is the only one you're seeing? It should cause similar errors when dumping traces or paths outside your open_basedir... Or is it just coincidence because of your weird default `DOCUMENT_ROOT`?

jnvsor:
Alrighty then, can you confirm if b2effe979abe59caf090b89b607fb0256e467139 on my fork fixes it for you @Bugophobia ? If so I'll tag it up and push it

ghost:
> Alrighty then, can you confirm if [b2effe9](https://github.com/kint-php/kint/commit/b2effe979abe59caf090b89b607fb0256e467139) on my fork fixes it for you @Bugophobia ? If so I'll tag it up and push it

It seems to work, since you made the warning silent. The warning in `init.php` was the only problem

PR #404, opened 2023-01-31
Composer 2.0 Compatibility

10n:
Composer 2.0 Compatibility: 
* packages are now wrapped into a "packages" top level key instead of the whole file being the package array

Documentation: https://getcomposer.org/upgrade/UPGRADE-2.0.md

### DIFF
--- a/src/Utils.php
+++ b/src/Utils.php
@@ -118,7 +118,9 @@ final class Utils
                     continue;
                 }
 
-                foreach ($packages as $package) {
+                // Composer 2.0 Compatibility: packages are now wrapped into a "packages" top level key instead of the whole file being the package array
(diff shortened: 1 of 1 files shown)


jnvsor:
Nicely spotted. Can you run `composer build` and commit the changed phar file, or should I do that for you?

10n:
> Nicely spotted. Can you run `composer build` and commit the changed phar file, or should I do that for you?

Hi, I would appreciate if you could run it, the commit was done in the web interface and locally I do not have all the dependencies to run the build.

Thanks

jnvsor:
Merged, tagged, and pushed. Thanks for spotting that!

PR #408, opened 2023-03-14
Optimize check for Composer extras

crocodele:
The Composer extras are only used if the flag constants aren't defined already. If the flags have been defined, we might as well skip searching for and parsing the composer.json files and thereby save some file system calls.


jnvsor:
Looks good to me, do you want to build the phar file with `composer build` or should I? (You may need to set `phar.readonly = Off` in your ini)

crocodele:
> Looks good to me, do you want to build the phar file with `composer build` or should I? (You may need to set `phar.readonly = Off` in your ini)

Thanks! I built the updated PHAR.

crocodele:
@jnvsor, I'm not sure why the build check fails. When I run the same commands locally, I don't get any errors from `git diff-files`. Can you help please? :slightly_smiling_face:

jnvsor:
That's strange. The phar build is supposed to be reproducible. I even strip it of timestamps for reproducibility.

Can you tell me what build environment you're using? I (and github actions) are using PHP 8.2 on linux, but I don't think the PHP version is to blame.

It doesn't seem to be a CRLF issue (Lots of `0xA4` replaced with `0xB4`) but maybe it's a file permission issue with NTFS if you're using windows?

I'll build it for you so we can get this merged. Would you mind helping me debug what caused the build issue?

jnvsor:
Merged in c19f79aa2a14bb1b7d02db750745d811cf2e600a - can we continue talking about the build in #409 ?

PR #411, opened 2023-03-30
fix `date() expects parameter 2 to be int, bool given`

totoprayogo1916:
image

### DIFF
--- a/src/Zval/Representation/SplFileInfoRepresentation.php
+++ b/src/Zval/Representation/SplFileInfoRepresentation.php
@@ -181,7 +181,7 @@ class SplFileInfoRepresentation extends Representation
 
     public function getMTime(): ?string
     {
-        if (null !== $this->mtime) {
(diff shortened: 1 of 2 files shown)


jnvsor:
Can you make a test case that triggers this? I'd rather have it covered (See also 1033fc4a5075e15b9b46be20c246069ab44bc243)

And if you can make a test case that triggers it it would be better to change 66 and make mtime a null in that case

jnvsor:
Well yeah if you set a property to an invalid value it's going to do strange stuff. `mtime` there should be a `?int` so `false` would be an invalid property. And we can't mark these properties `readonly` because we support down to 7.1

Can you show an example with a real `SplFileInfo` that has both `getPathname` and `getRealPath` but `getMtime` returns `false`? Is this an edgecase that only occurs on certain filesystems? Where did you run into this bug?

totoprayogo1916:
I'll fix it some other time.
I tried again, but I forgot how to return the false value.

PR #416, opened 2023-08-31
Add 'dev' keyword to composer.json to trigger `--dev` prompt

Ayesh:
Composer 2.4 and later can automatically suggest to use the `--dev` flag when someone runs `composer require kint-php/kint` without the `--dev` flag.

See [Get Composer to suggest dev packages to `require-dev`](https://php.watch/articles/composer-prompt-require-dev-dev-packages)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,7 +1,7 @@
 {
     "name": "kint-php/kint",
     "description": "Kint - debugging tool for PHP developers",
-    "keywords": ["kint", "php", "debug"],
+    "keywords": ["kint", "php", "debug", "dev"],
     "type": "library",
     "homepage": "https://kint-php.github.io/kint/",
     "license": "MIT",


jnvsor:
There's a risk that someone leaves in a kint call and only notices when they put it live and production breaks with a call to a non-existent function, while it works fine on the dev machine.

This is in contrast to things like phpunit or psalm - unless your application is some sort of unit testing service it's probably never going to enter the phpunit codebase during production.

We also advise to use `Kint::$enabled_mode` to disable it in production in the docs for the same reason.

This is probably why laravel includes `var-dumper` in it's `require` not it's `require-dev` and why symfony doesn't add a `dev` keyword to `var-dumper`

I don't think we should automagically advise people to do something that could break production in unexpected ways. Let them do it themselves if they understand what they're doing

Ayesh:
Absolutely makes sense. Thanks for your quick review 🙏.

PR #417, opened 2023-10-03
add aante-dark theme

LoranRendel:
This commit adds aante-dark theme based on aante-light with inverted colors.


jnvsor:
I see a few visibility issues here - could you fix these for me?

* The folder background has gone from `argb(#FFF, 90%)` to `argb(#000, 10%)` and the lack of opacity makes it hard to read. It should probably just stay white. Remove the `$backdrop-color` change
* The minitrace is illegible because the text is white on grey. Solarized dark had to fix this too (It's at the end if you want to copy it)

Lastly, if it's really just a color inversion then we should be able to base it off the existing aante-light. Didn't do this for solarized since it has very little custom css (Just tweaking variables) but should probably do it for aante since it's much heavier in that way.

Please change the variable assignments in aante-light to `!default` ones and base aante-dark off it, then there should be much less code. If you don't want to I'll do it myself after merging

LoranRendel:
My mistake, I've only tested it on a black body.

`$backdrop-color: rgba(#000, 0.9)` will be better.

I'll try to make all changes in next 24h.

LoranRendel:
I've made changes.

Also I think changing `margin` to `padding` in `.kint-rich` may be good idea to avoid stripes in all themes:

before/after
imageimage

jnvsor:
Could you run a `composer format` and `composer build` and that should do it

jnvsor:
Can you get rid of the margin/padding thing and do a build?

Padding looks fine if you're dumping multiple at once or have a page only for dumps but if you're dumping at different places on a page the margin looks much cleaner, so I'd like to keep that as the default

LoranRendel:
Yes, I will revert the commit but in contrast theme it doesn't look good.

jnvsor:
Well presumably you'd be using the dark theme on a dark site and it would be less jarring, but since we have no control over where it's run I'd rather keep it as margin. Can I just add a `composer build` onto fc86954 and merge this then?

LoranRendel:
Yes, of course.

This commit already has `composer build`.

jnvsor:
> This commit already has `composer build`.

Nope, the phar file hasn't been changed - You may need to set `phar.readonly = Off` in your ini if you want to build it, but I'll just do it now since I have it all set up

LoranRendel:
Fixed in `2de8ac9`

jnvsor:
Merged to master - Since there are no bugs right now I'll tag it in the next minor version closer to PHP 8.3 release

PR #429, opened 2024-10-28
Migrate SCSS variables to CSS3 to enable auto dark and more features

### LINKED ISSUE #428: New color scheme with hybrid dark mode support
> Originally discussed in: https://github.com/kint-php/kint/issues/422#issuecomment-2381454524

In Kint 6, we can drop IE support and use [CSS variables](https://www.google.com/search?q=css+variables&ie=UTF-8&hl=en) together with media queries to detect dark mode support by the device or the browser and automatically apply a cool looking dark variant of that theme:

```css
@media (prefers-color-scheme: dark) {...}
```

This preferably needs to be applied to all themes currently residing in the resources directory.

While at it, the color scheme can be refreshed for a more modern and streamlined look, with more contrast.

DRSDavidSoft:
### Description
This is a preliminary PR to address #428. Changes so far include:

- Converted all SCSS variables to CSS variables instead where required (no scss variable remained after this change)

- Drop IE support (eliminating the need for base64 encoding of SVG images, reducing both file clutter and size)

- General cleanup to improve consistency across themes

**Note:** This PR currently _doesn't_ introduce visual changes. If there are any differences from the previous styles, please flag them. I’ve reviewed this but would appreciate a thorough check.

**Additional Requests:**
- I used `d($_SERVER);` to verify theme consistency. But it might not output use of all the Kint features, so could you please prepare a script for me that would make use of all the Kint features for a detailed comparison between the previous and old styles?

- For theme files, we currently have separate files for `solarized-light` and `solarized-dark`. Ideally, we should consolidate to a single `solarized.css` file with color variables adaptable to both themes. Users might prefer explicitly setting light or dark themes, though, and we should avoid disrupting existing user settings. I propose a “theme alias” feature in Kint, mapping old themes to a unified auto theme while applying `kint-dark` or `kint-light` classes as appropriate. This would allow us to ship just one file instead of three. Let me know if you agree, and I’ll work on a PR.

- Once decisions are finalized, another PR could be opened to refresh theme colors using professionally curated palettes in preparation for the Kint 6 release.

### Theme Variable Enhancements
- Added CSS variables for spacing, colors, and design elements for easier customization and theme management.

### Simplification and Consistency
- Replaced hardcoded values with CSS variables in order to enable declarative overriding (e.g. using `@media (prefers-color-scheme: dark)` query)
- Simplified complex selectors and properties to improve readability and maintainability.

The changes will help in maintaining a consistent design and make future modifications more straightforward.

### Tests
Verified the produced `css` in the compiled directory against the older versions by setting the following:

```php
RichRenderer::$theme = 'aante.css';
```

To ensure new styles exactly match previous themes.


jnvsor:
Good work

> I propose adding a "theme alias" feature in Kint which is very simple, it would map the old theme files to the new auto theme, while also applying a kint-dark or kint-light class to the renderer.

Hmm I'd rather apply it to the body since there's only supposed to be "1 true stylesheet" for kint in a page. Are there any fancy frontend systems you know of that could break that? (ie. react)

I'll review this today

resources/sass/aante.scss
```diff
@@ -0,0 +1,169 @@
...
+
+  --border-color: #d7d7d7;
+  --border-color-hover: #aaa;
+
+  --alternative-background: #fff;
+  --highlight-color: #cfc;
+
+  --caret-image: url("data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 30 150'><path d='M6 7h18l-9 15zm0 30h18l-9 15zm0 45h18l-9-15zm0 30h18l-9-15zm0 12l18 18m-18 0l18-18' fill='%23555'/><path d='M6 126l18 18m-18 0l18-18' stroke-width='2' stroke='%23555'/></svg>");
+
+  @media (prefers-color-scheme: dark) {
```
jnvsor: Let's add the media tag in a separate commit afterwards

resources/sass/aante.scss
jnvsor, line 1: Move `aante` back to `aante-light` for now (Makes diffing easier)

resources/sass/_base.scss
```diff
@@ -1,47 +1,47 @@
 // Variables for themes to override
-$spacing: 4 !default;
-$color_size: 16 !default;
+:root {
```
jnvsor: I'm concerned the variables might clobber user themes with common names like `--spacing`. We could namespace everything but I think a simpler solution would be to attach them to `.kint-rich` rather than `:root` since we control everything that should be in there anyway

resources/sass/_base.scss
```diff
@@ -1,47 +1,47 @@
...
+  --spacing: 4px;
+  --color-size: 16px;
 
-$backdrop-color: rgba(#fff, 0.9) !default;
-$main-background: #e0eaef !default;
-$secondary-background: #c1d4df !default;
+  --backdrop-color: rgba(255, 255, 255, 0.9);
```
jnvsor: Is there a technical reason for changing this? Are we heading towards a future without sass?

resources/sass/_base.scss
```diff
@@ -51,18 +51,16 @@ $caret-image: url("data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5
     }
   }
 
-  &::selection,
-  &::-moz-selection,
-  &::-webkit-selection {
-    background: $border-color-hover;
-    color: $text-color;
+  &::selection {
```
jnvsor, line 52: Could you split this out into its own commit?

resources/sass/_base.scss
```diff
@@ -1,47 +1,47 @@
 // Variables for themes to override
-$spacing: 4 !default;
-$color_size: 16 !default;
+:root {
+  --spacing: 4px;
```
jnvsor, line 3: I was worried this was without a unit for a reason but it looks like everywhere `$spacing` and `$color_size` were used used pixels

resources/sass/_base.scss
```diff
@@ -165,77 +163,77 @@ $caret-image: url("data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5
...
     border-bottom-width: 0;
-    padding: $spacing * 1px;
+    padding: var(--spacing);
     float: right !important;
-    margin: ($spacing * -1px) 0;
-    color: $variable-name-color;
-    background: $secondary-background;
-    height: $color_size * 1px + $spacing * 2px;
-    width: $color_size * 10px;
+    margin: var(--spacing) 0;
```
jnvsor: This is flipped

resources/sass/_base.scss
```diff
@@ -379,25 +379,27 @@ $caret-image: url("data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5
...
-    border: $border;
+    border: var(--border);
     background: {
       color: #ccc;
       image: url('data:image/svg+xml;utf8,<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 2 2"><path fill="%23FFF" d="M0 0h1v2h1V1H0z"/></svg>');
-      size: if($color_size >= 30, 30px, 100%);
+      size: 100%;
     }
+    max-width: 30px;
+    max-height: 30px;
```
jnvsor: Could you move these up above `background`? (Also nice cleanup btw)

resources/sass/aante-dark.scss
```diff
@@ -1,23 +1,24 @@
...
 
-$main-background: invert(#f8f8f8);
-$secondary-background: invert(#f8f8f8);
+@import "aante";
 
-$variable-type-color: invert(#06f);
-$variable-type-color-hover: invert(#f00);
+:root {
+  --main-background: #070707;
+  --secondary-background: #070707;
```
jnvsor: No technical reason for getting rid of the invert?

resources/sass/aante.scss
```diff
@@ -0,0 +1,169 @@
...
+      margin-top: -1px;
+
+      li,
+      li + li {
+        margin: 0 0 0 4px;
+      }
+
+      li {
+        border-bottom-width: 0;
+        height: calc(var(--spacing) * 6);
```
jnvsor: The original was `spacing * 6px + 1px` - there's a gap in the border (On firefox at least) without the extra pixel

resources/sass/original.scss
```diff
@@ -1,27 +1,30 @@
 @import "base";
 
+:root {
```
jnvsor: Since these are specific to `original` I deliberately didn't bother putting them in variables. Maybe 1 and 2 should be since they're used more than once...

In that case what do you think of naming them `--gradient-start`/`--gradient-end`?

resources/sass/_base.scss
```diff
@@ -165,77 +163,77 @@ $caret-image: url("data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5
...
+      color: var(--text-color);
+      background: var(--main-background);
     }
   }
 
   // The &telrec; icon is slightly smaller than the others,
   // so make it a bit bigger to line it out better
   .kint-search-trigger {
-    font-size: $color_size * 1.25 * 1px;
+    font-size: calc(--color-size * 1.25);
```
jnvsor: Broken. Missing `var()`

All of this stuff was carefully tweaked numbers while checking browsers but it looks like `calc` behaves differently and it'll need to be straightened out afterwards

resources/sass/_base.scss
```diff
@@ -165,77 +163,77 @@ $caret-image: url("data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5
...
     float: right !important;
     cursor: pointer;
-    color: $main-background;
+    color: var(--main-background);
     position: relative;
     overflow: hidden;
 
     // Carefully balanced on multiple browsers
-    line-height: $color_size * 1.1 * 1px;
+    line-height: calc(--color-size * 1.1);
```
jnvsor: Broken. Missing `var()`

jnvsor (changes requested):
Additionally a weird rendering artifact only occurs on firefox not chrome:

image

I'll try to debug this myself

resources/sass/_base.scss
jnvsor: In my tests you can just set this to `var(--color-size)` and it will work now

resources/sass/_base.scss
DRSDavidSoft: Awesome, that's actually one point I forgot to mention. I was initially going to namespace the vars like `--kint-spacing`, but attaching them to `.kint-rich` makes sense too

resources/sass/_base.scss
DRSDavidSoft: The technical reason for the `--backdrop-color` change is that it needs to be overridable from other themes.

The reason behind changing `rgba(#fff, 0.9)` to `rgba(255, 255, 255, 0.9)` is because the former doesn't appear to be valid CSS3 color, while the latter is.  We could still use Sass-style colors, but it needs to be transpiled into regular CSS for it to work and I couldn't figure it out in the spur of the moment.

Are we heading towards a future without Sass? No, it doesn't make sense to stop using all the nice benefits of Sass, such as nested declarations. The intention is to only support using native CSS variables, so that they can be overridden in run time (rather than build time) by other themes, depending on the situation (e.g. using `media` queries), which can't be done using Sass-only variables.

resources/sass/_base.scss
DRSDavidSoft: Sure, made this:

https://github.com/kint-php/kint/pull/430

I assumed it would be fine to integrate it here at first, in order to reduce clutter and avoid having to open multiple PRs.

resources/sass/_base.scss
DRSDavidSoft: Right? It was also my initial assumption, however upon further look it appears that this was just done to make use of Sass calculation.

It makes sense to explicitly set the unit to `px` here rather than everywhere else in the CSS rewrite, correct? Also, a theme might use another unit of their choosing.

Using CSS' native `calc()` doesn't care _what_ the initial unit is, it just performs multiplication or division on it, which is our goal here.

resources/sass/_base.scss
DRSDavidSoft: Good catch! Sorry for this, fixed in the newer commit that I will push soon.

resources/sass/_base.scss
DRSDavidSoft: Thanks! Sure, done! Will push soon.

resources/sass/aante-dark.scss
DRSDavidSoft: I wouldn't say a "technical" reason, but here are the reasons:
1. `invert(...)` is a Sass feature, not native to CSS. I could not figure out how to apply the colors so that it would transpile correctly.
2. In the upcoming commits, I plan to "refresh" the colors that is used. When @LoranRendel helped add `aante-dark` last year, they used the inverted colors, which is fine, but if we used "dark-optimized" values (that are not just the `invert()`-ed variant of the light ones), it can look better.

This means that if the chosen values are completely unrelated to those of the light variant, we can forgo using `invert()` altogether and just use custom dark-optimized values that look nicer.

resources/sass/aante.scss
DRSDavidSoft: Man, I am sorry I don't know why I missed these things, very important catch, thanks!

Fixed in the newer commit which I will push soon.

resources/sass/original.scss
DRSDavidSoft: Exactly, I just named them `color1` and `color2` as a temporary placeholder.

I prefer to use `gradient-start` and `gradient-end`, just that since there are two gradients used here, maybe they can be named `primary-gradient-[start|end]` and `secondary-gradient-[start|end]`, I just didn't know which was which when I did it.

Alternatively, would be to make the entire gradient a variable instead. Then, instead of using CSS to apply them to the elements in question, the newly created `--background-image` variable can be used instead.

resources/sass/_base.scss
jnvsor: The `rgba(#fff, 0.9)` _is_ sass and it should compile to `rgba(255, 255, 255, 0.9)` automatically. I'm not sure why it won't, maybe sass explicitly disables it in css variables?

resources/sass/original.scss
jnvsor: There are 2 gradients but `color3` `color4` and `highlight-color` are only used once so you might as well make them literals

resources/sass/_base.scss
jnvsor: I think the intent was to force the unit to be pixels (Since if you use something like em in a recursive structure you'll get ballooning sizes)

Also for line-height there's merit to unitless input (`line-height: 1.1` is roughly `line-height: calc(font-size * 1.1)`) but again we don't actually use it. I have no strong opinions one way or the other

resources/sass/_base.scss
jnvsor: Ugh. Got it working but the syntax sucks:

```scss
--backdrop-color: #{rgba(#FFF, 0.9)};
```

resources/sass/_base.scss
DRSDavidSoft: I must apologize again, I really don't know why I missed some of these 😅 

Anyways, piece of advice (~15 years of web dev) is that using this approach is not a sure thing across all the different screens, DPI scalings, fonts, OSes, etc.

Can you please include of how it is supposed to look? I believe we can include a SVG icon that would look way better, anyways.

resources/sass/_base.scss
DRSDavidSoft: I believe using a `flex` solution is better than relying on `line-height`, that's an outdated method.

resources/sass/_base.scss
jnvsor: The problem is that we have 3 icons on the right:

* ⌕
* ↧
* ⇄

I don't know about your screen but on mine one of these is much smaller than the rest:

> image

It could just be my font but 1.25 has been pretty consistent for me

DRSDavidSoft:
> Hmm I'd rather apply it to the body since there's only supposed to be "1 true stylesheet" for kint in a page. 

Good call, so setting the theme to an alias `aante-dark` will actually load the `aante` theme, and apply `kint-dark` class to the body itself

> Are there any fancy frontend systems you know of that could break that? (ie. react)
I don't think there would be a problem, although I don't think React would be used this way with PHP.

If a request from React is made to a PHP backend (e.g. Using Axiom), then it's most likely a "fetch" type request (à la ajax) and won't be directly rendered in the page.

There is one problem that comes to mind, which is that we will need to declare our styles as `.kint-dark .kint-rich { ... }` instead of `.kint-rich.kint-dark { ... }`. Actually, not a real problem, just something to have in mind.

DRSDavidSoft:
Please review the requested changes and mark each conversation that is correctly resolved as such, if it wasn't resolved correctly please add a new reply.

Thanks! 🤝

jnvsor:
> Good call, so setting the theme to an alias aante-dark will actually load the aante theme, and apply kint-dark class to the body itself

> If a request from React is made to a PHP backend (e.g. Using Axiom), then it's most likely a "fetch" type request (à la ajax) and won't be directly rendered in the page.

Yeah but you can't set a class on the body from kint in PHP, would have to be in JS. I can imagine different dumps with the light and dark version toggling the class on the body which is not what we want I presume.

Maybe it _should_ be a class like `.kint-rich.kint-dark` but that wouldn't work for the folder... I guess we could just have JS store the dark flag of first dump moved to the folder and use that for the folder class

resources/sass/aante.scss
DRSDavidSoft: Sure, I will keep it for now in this WIP PR, then I will move it to its own PR before merging this one.

resources/sass/_base.scss
DRSDavidSoft: Excellent, and I agree the syntax is less than visually appealing, but where we _need_ to make use of Sass utils, it's acceptable.

I would say that after applying a "color refresh" to make it look better, some of these Sass definitions wouldn't be used as we will just apply an absolute CSS color.

resources/sass/_base.scss
jnvsor: Actually it could just be `line-height: 1` since we set the font-size to `--color-size` anyway (sorry for nitpick heh)

resources/sass/_base.scss
DRSDavidSoft: I agree, also there is the possibility of not using `line-height` for this and make use of other approaches.

jnvsor: The dom structure needs a rework anyway (It's too deep with too many elements and makes the browser take longer to render the dom than kint takes to render the markup) but that's a big change that touches everything including JS so it's definitely not on the table rn.

resources/sass/_base.scss
DRSDavidSoft: I am not applying any custom fonts or anything like that, but the magnifying glass on my screen clearly has the same width as other icons, while in your screenshot it does not:

image

I vote to use SVG icons here too. If you accept, I will whip a quick PR that uses Sass to load SVG icons directly into the CSS code.

One advantage this has is better theming support, like how the caret is themed.

Not lying, I disliked how these icons looked right from start when they were introduced, they don't "feel" like the rest of Kint, and I believe they can be improved if SVG icons are used instead.

jnvsor: Alrighty, whip me up an SVG then. I wrote the one in the caret by hand years ago.

I know you can make an svg spritesheet where each icon is on its own "layer" but I don't remember how to do it.

As far as the icons go other than the search icon (Which is pretty universally a magnifying glass) I have no particular preference. The ones in use were just easily usable as htmlentities.

DRSDavidSoft:
> Maybe it _should_ be a class like `.kint-rich.kint-dark` but that wouldn't work for the folder... I guess we could just have JS store the dark flag of first dump moved to the folder and use that for the folder class

If I am not mistaken, the folder functionality already makes use of JS, right? If so, it makes sense for PHP to set the `kint-dark` class on the `kint-rich` element, then _when_ the folder feature is initialized, using JS we will check whether this class if set, and if so, copy it to the folder element.

> I can imagine different dumps with the light and dark version toggling the class on the body which is not what we want I presume.

I wouldn't say that it would make sense, but I believe _if_ the user can set the theme to be anything for each individual dump, it would make sense that they could also turn the dark mode on/off by specifying another "theme alias".

In conclusion, this makes more sense to me when I think about it, rather than setting `kint-dark` on the body itself.

> you can't set a class on the body from kint in PHP, would have to be in JS.

Quite right, the closest thing I can imagine is having a parent `.kint-dark` element, but none of this makes sense as opposed to just applying the `kint-dark` class directly on `kint-rich` element.

I guess _functionality speaking_, it doesn't make sense to have per-dump dark control, so a "global" dark mode control approach would've been nice. But _technically speaking_, setting it on the whole page doesn't make sense, and setting it on each individual dump makes more sense.

jnvsor:
> If I am not mistaken, the folder functionality already makes use of JS, right? If so, it makes sense for PHP to set the kint-dark class on the kint-rich element, then when the folder feature is initialized, using JS we will check whether this class if set, and if so, copy it to the folder element.

Yeah that's exactly what I meant. Let's go with that (I can work on the php/js side) once this migration is done

resources/sass/_base.scss
```diff
@@ -1,47 +1,47 @@
 // Variables for themes to override
-$spacing: 4 !default;
-$color_size: 16 !default;
+:root {
+  --spacing: 4px;
+  --color-size: 16px;
```
jnvsor, line 4: This was originally intended for color swatches but ended up getting used for right side buttons too. It's basically the maximum height of blocks inside the bar. It should probably be renamed in a patch _after_ this one is live

resources/sass/_base.scss
```diff
@@ -165,77 +163,78 @@ $caret-image: url("data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5
...
     float: right !important;
     cursor: pointer;
-    color: $main-background;
+    color: var(--main-background);
     position: relative;
     overflow: hidden;
-
-    // Carefully balanced on multiple browsers
-    line-height: $color_size * 1.1 * 1px;
+    display: flex;
```
jnvsor, line 208: I know with large dumps the reflows can get nasty. Could this have perf impacts?

resources/sass/_base.scss
jnvsor: For now I'll resolve this since it's equivalent to the old styles

resources/sass/_base.scss
DRSDavidSoft: Sure, I set it to `lint-height: normal` and also applied the flex to make sure it is properly centered, check the new commit to see

(Also don't worry, I have the exact same feelings regarding the nitpicks, even that 1px makes my OCD go creazy 😂 so it'll be awesome to make sure it looks fine without any pixel displacements, whatsoever)

jnvsor: > Depends on the user agent. Desktop browsers (including Firefox) use a default value of roughly 1.2, depending on the element's font-family

Per mdn: https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#normal

That's a lot of conditions. `line-height: 1` is probably safer (Until your svg patch comes in anyway)

jnvsor (changes requested):
Sorry I didn't press the big green button

jnvsor:
@DRSDavidSoft any movement on this? 8.4 is now scheduled to release on the 21st so we have a little under 2 weeks.

I'd like to push an RC by the end of this weekend, but I still have more changes to make before then.

If you can't make it by then just take your time and we'll aim for 6.1

DRSDavidSoft:
@jnvsor I was just a bit busy at work, if possible I would like to focus the PR only at migrating the variables for now to keep it simpler and allow us the time for it to be merged in `6.0`. Then, I will work on another PR to address the auto dark mode and the new icons. I hope that's also fine with you.

If so, can you please make a list of the changes that I need to apply for this PR to be merged? That would help me prioritize the changes as well as allowing #428 to be implemented in a later PR that will be based on this one.

jnvsor:
Yeah that works for me

I'll just resolve everything and make a new review then

resources/sass/aante-dark.scss
```diff
@@ -1,23 +1,24 @@
 // $author Ante Aljinovic https://github.com/aljinovic
 // modified dark version
 
-$main-background: invert(#f8f8f8);
-$secondary-background: invert(#f8f8f8);
+@import "aante-light";
 
-$variable-type-color: invert(#06f);
-$variable-type-color-hover: invert(#f00);
+:root {
```
jnvsor: Put vars in .kint-rich (In all scss files)

In most children extending _base they already have a .kint-rich block so merge them rather than duplicating them

resources/sass/aante-light.scss
```diff
@@ -1,25 +1,46 @@
...
+  --border-color: #d7d7d7;
+  --border-color-hover: #aaa;
 
-@import "base";
+  --alternative-background: #fff;
+  --highlight-color: #cfc;
+
+  --caret-image: url("data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 30 150'><path d='M6 7h18l-9 15zm0 30h18l-9 15zm0 45h18l-9-15zm0 30h18l-9-15zm0 12l18 18m-18 0l18-18' fill='%23555'/><path d='M6 126l18 18m-18 0l18-18' stroke-width='2' stroke='%23555'/></svg>");
+
+  @media (prefers-color-scheme: dark) {
```
jnvsor: Remove this for now, we'll add it later

resources/sass/original.scss
```diff
@@ -1,27 +1,30 @@
...
+  --kint-color1: #e3ecf0;
```
jnvsor: Change `--kint-color1` to `--gradient-start` and `kint-color2` to `--gradient-end`

Remove the other 3 vars. Since they're only used once just use literals

resources/sass/_base.scss
jnvsor: I'll perf check this myself

resources/sass/_base.scss
```diff
@@ -379,25 +380,27 @@ $caret-image: url("data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5
...
     margin-left: 10px;
-    border: $border;
+    border: var(--border);
+    max-width: 30px;
+    max-height: 30px;
     background: {
       color: #ccc;
       image: url('data:image/svg+xml;utf8,<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 2 2"><path fill="%23FFF" d="M0 0h1v2h1V1H0z"/></svg>');
-      size: if($color_size >= 30, 30px, 100%);
+      size: 100%;
```
jnvsor, line 397: I'm not sure these changes won't break some layouts but I'll merge them first and test them myself later

jnvsor (changes requested):

resources/sass/_base.scss
DRSDavidSoft: I don't see why the new approach would break anything, but we could also go this route as well:
`size: min(30px, 100%)`
which has the same effect.

Alternatively, if you found a breaking change with either of these CSS-only implementations, we can revert back to the SCSS solution.

resources/sass/original.scss
DRSDavidSoft: Let's keep these as variables. My reasoning is that when we need to introduce the dark edition of this theme, we need the vars so that we could override them.

If we don't use variables here now, in the next PR those needs to be added later anyway.

jnvsor: Good point... Then just leave them unnamed for now, I'll think of something to call them later

resources/sass/aante-dark.scss
DRSDavidSoft: I applied this change, although I believe there could be some benefits for using the `:root` space, particularly if we use a short and concise prefix such as `--kt-`.

Would you prefer the following declarations? Otherwise, I'll put the variables in `.kint-rich` as requested.
```scss
:root {
  --kt-variable-type-color: ...
  --kt-variable-type-color-hover: ...
}
```

This would allow the users to easily override Kint themes using custom values later on.

resources/sass/original.scss
DRSDavidSoft: The main purpose for using vars instead of literals is being able to override them in the _TBA_ dark theme variant of this theme.

resources/sass/aante-dark.scss
jnvsor: Nah just leave it in `.kint-rich` - it's just as hard for a user to write

```css
.kint-rich {
  --variable-type-color: #F00;
}
```

As

```css
:root {
  --kt--variable-type-color: #F00;
}
```

jnvsor:
I'm off to work now, I'll review tweak and merge later today

resources/sass/aante-dark.scss
DRSDavidSoft: Actually what I mean is the ability to override the variable, but sure, I submitted the PR using the `.kint-rich` instead of `:root`

Note: using the `.kint-rich` approach, the user have to supply a more qualifying css selector in order to override their variables. e.g. add `kint-custom` class to the body, then insert overriding vars as `.kint-custom .kint-rich`, hopefully I was able to demonstrate my point.

jnvsor: You can also do this with `:root .kint-rich` without touching the HTML, but even with that it's not a downgrade.

Before you had to style individual elements since you didn't have access to variables at all. Now you can still do it without more specific selectors if you make sure your css comes after kints (Which is admittedly pretty hard to do but easy enough for actual kint themes)

jnvsor:
There's a few small things but I'll just take care of them here. For now could you hold off on making any more changes until I've pushed? Otherwise we're just going to cause conflicts

jnvsor (approved):

jnvsor:
Merged in 63fcf330cd07bcd139230bb0354812d65260a029 with some tweaks after that.

The CSS could really use some love. I noticed the `.kint-parent` doesn't have a fixed height it's just flowing with the content. That could potentially be causing client-side perf issues.

Anyway I'll leave CSS here and push rc1 when I'm done with my changes

PR #437, opened 2025-06-20
[GH Pages] Update arrows icon to crosshair

igorsantos07:
I guess this was an old icon that wasn't noticed


jnvsor:
@igorsantos07 You're right, well spotted. The folder icon is wrong too. I'll copy the SVGs to the site and use them instead since there are no real equivalents.

jnvsor:
Fixed in 79d9734 - thanks @igorsantos07

PR #439, opened 2025-08-15
refactor: setAccessible no longer effect since PHP 8.1

ddevsr:
> As of PHP 8.1.0, calling Reflection*::setAccessible() [has no effect](https://www.php.net/manual/en/reflectionproperty.setaccessible.php); all properties are accessible by default. Currently, there is a [RFC](https://wiki.php.net/rfc/deprecations_php_8_5#deprecate_reflectionsetaccessible) which proposes to issue a deprecation warning starting with Php 8.5.

cc @jnvsor


jnvsor:
Hi @ddevsr, can you use the `KINT_PHP_81` constant instead and squash it into a single commit? Thanks!

ddevsr:
You mean new constant `KINT_PHP_81` in init.php?

Edit: Ohh ok use `KINT_PHP81 === false`

jnvsor:
I'd use `!KINT_PHP81` but no real difference.

Could you run `composer format` and `composer build` and commit the changes? That way the phar file will be reproducible and CI will go green :)

jnvsor:
I see what's happening, php-cs-fixer is reordering some docblocks and that's making the build fail. Can you run a `composer install` and explicitly use the php-cs-fixer version in the lockfile? `PHP_CS_FIXER_IGNORE_ENV=1 ./vendor/bin/php-cs-fixer fix` should set things straight before the build

ddevsr:
Push unecessary file in this PR?

jnvsor:
Nah, if you have the php-cs-fixer version in the lockfile (So just a `composer install` with the version in that commit) it won't make those changes and the build will succeed

If you don't want to `composer install` for some reason you could checkout other changes before the format and build from there and it should work too

**Edit:** If that's too much I could rebuild it for you and push the PR manually

jnvsor:
Merged as 33e535cd3ec8d2d510cbea544e4b093e10930c3d

ddevsr:
@jnvsor Tagged new version for support PHP 8.5 soon?

jnvsor:
@ddevsr before php8.5 release yes, but because kint is very sensitive to very niche parts of PHP I like to leave it as late as possible before release since even with a feature freeze there can be breaking changes in the last week

**Edit:** but if you're concerned about the version constraint in codeigniter don't worry, I don't put upper version limits on kint so it _probably_ won't break anything to just leave it on the current version so long as you're not committing the lockfile or update the version when updating to 8.5

PR #440, opened 2025-08-16
chore: allow unsupported php version & parallel config in `PHP-CS-Fixer`

ddevsr:
- Bump `friendsofphp/php-cs-fixer` to minimum `^3.86.0`
- `PHP_CS_FIXER_IGNORE_ENV` already deprecated and change to new config
- Set parallel config for faster fix


jnvsor:
I usually update this once a year, with the specific version in the lockfile. It typically requires either:

1. An updated config to stop it from changing a bunch of code
2. A rebuild because it changed some stuff anyway

In this case I'll want to make decisions myself, but thanks for the start anyway

jnvsor:
Merged (With other updates) in bd5136c

